PR #5, opened 2021-01-31
hardis:auth:login

nvuillam:
- New command force:auth:login 
- manage login using JWT for CI


nvuillam:
- New command force:auth:login
- manage login using JWT for CI

PR #203, opened 2022-02-18
Add installationkey flag on package install

clementfernandez:
- hardis:package:install
  - Add -k, --installationkey CLI param and prompts user for it if not supplied


nvuillam (approved):
good :)

PR #206, opened 2022-03-08
Ignore missing Apex test when project is empty

clementfernandez:
This allows the CI/CD to be setup when no Apex code/classes have been written yet.

We catch the error string and handle french and english languages so far.


nvuillam (changes requested):
And the changelog ? ;)

PR #283, opened 2022-10-18
Fixes/findduplicates documentation

mkp0x:
I don't understand why documentation for old commands is still generated : 

`sfdx hardis:doc:plugin:generate` keeps generating `docs/hardis/manifest/duplicate/{find,remove}.md` altough there seems to be no reference to these commands.

`mkdocs.yml` also gets updated after I run sfdx hardis:doc:plugin:generate and it keeps including these commands 

```yaml
        manifest:
          duplicate:
            find: hardis/manifest/duplicate/find.md
            remove: hardis/manifest/duplicate/remove.md
```

I must be doing wrong on the documentation generation but I haven't found what


nvuillam:
@mkp0x remove lib folder and tsc again :)
Probably old .js from .ts are still here :)

nvuillam (approved):
good :)

PR #284, opened 2022-11-08
Use new `force:package:beta:install`

mkp0x:
## [2.96.0] 2022-11-08

- Replace `sfdx force:package:install` with `sfdx force:package:beta:install`
- Do not cause deployment to fail when a deploying an older managed package version
  - Instead, deployment will assume the newer version meets the requirement


.trivyignore
```diff
@@ -24,3 +24,4 @@ CVE-2022-33987
 CVE-2022-37616
 DS001
 DS002
+CVE-2022-3517
```
nvuillam, line 27: et l ordre alphabétique, y est l'chien qu'l'a bouffé ? :D

nvuillam (approved):
Great, thanks for your contribution :)

nvuillam:
thanks @mkp0x , now you can create the release :)

PR #287, opened 2022-11-24
Authorizations linter

nvuillam:


src/commands/hardis/lint/access.ts
```diff
@@ -57,40 +57,40 @@ export default class Access extends SfdxCommand {
   protected static sourceElements = [
     {
       regex: `/**/*.cls`,
-      type : 'Apex classes',
```
nvuillam: @mickaelgudin please use type names of package.xml: ApexClass, CustomField, CustomObject... :)

nvuillam (commented):
Plz check comments :)

nvuillam:
@mickaelgudin you also have MegaLinter issues :)
Either some words you used are right but unknown to cspell (in that case update .cspell.json) , either you made typos that must be corrected ^^

PR #302, opened 2023-01-31
Fix clean hidden items

clemfernandez:


nvuillam (approved):
Elegant fix @clemfernandez :)

PR #338, opened 2023-03-27
WIP: Allow to deploy in delta

nvuillam:


src/common/utils/gitUtils.ts
```diff
@@ -32,3 +32,18 @@ export async function selectTargetBranch(options: { message?: string } = {}) {
...
+    `sfdx sgd:source:delta --from "${from}" --to "${to}" --output ${outputDir}` +
+    (sgdHelp.includes("--ignore-whitespace") ? " --ignore-whitespace" : "");
+  const gitDeltaCommandRes = await execSfdxJson(packageXmlGitDeltaCommand, this, {
+    output: true,
+    fail: false,
+    debug: options?.debugMode || false,
+    cwd: await getGitRepoRoot(),
+  });
+  return gitDeltaCommandRes;
+}
```
scolladon, line 49: **suggestion:** I think it could be possible to use the module directly

SGD is also [exposed as a module](https://github.com/scolladon/sfdx-git-delta#use-the-module-in-your-own-node-application):
```js
const sgd = require('sfdx-git-delta')

const work = await sgd({
  to: '', // commit sha to where the diff is done. [default : "HEAD"]
  from: '', // (required) commit sha from where the diff is done. [default : git rev-list --max-parents=0 HEAD]
  output: '', // source package specific output. [default : "./output"]
  apiVersion: '', // salesforce API version. [default : latest]
  repo: '', // git repository location. [default : "."]
})

console.log(JSON.stringify(work))
/* {
 *   config: config,
 *   diffs: { package: {...}, destructiveChanges: {...} },
 *   warnings: []
 * }
 */
```

nvuillam: As sfdx-git-delta can be used directly from the command line, I prefer to keep it as an external dependency, that way we are sure that users are not using one version of git-delta within sfdx-hardis and another locally installed as sfdx plugin on their computer :)

PR #423, opened 2023-09-18
Fixed code coverage comparison bug

legetz:
Previous code used string comparison, now new code uses float comparison


nvuillam (approved):
Seems good :)
Please can u test with beta (generated in a few minutes) before I release ? :)

PR #429, opened 2023-09-26
Fixed PR coverage to use float over string

legetz:
Fixed the issue of passing min coverage string into `updatePullRequestResultCoverage`. Instead pass the float `minCoverage` and `orgCoverage` in order to perform proper comparison of given numbers.


nvuillam (approved):
Thanks leo :)
next time please think to update CHANGELOG.md :)

PR #498, opened 2023-11-24
New feature to identify missing description on custom field

maximeg44:


nvuillam:
@maximeg44 great :)

I'll add the feature allowing to disable some checks before merging ^^

PR #554, opened 2024-02-02
Added bash to support project level scripts within Hardis image

legetz:
My customer project has few bash scripts which I will use in CI/CD steps. Using official SFDX Hardis image, hence want to have bash support


nvuillam:
@legetz thanks for the PR :)

Released in 4.26.3 :) https://github.com/hardisgroupcom/sfdx-hardis/releases/tag/v4.26.3

PR #577, opened 2024-04-02
Remove typo from package-no-overwrite.xml

clemfernandez:
Remove extra 's' from package-no-overwrite.xml

This incorrectly caused named credentials and remote site settings to be overwritten even though it looked like they were not


nvuillam (approved):
Thanks a lot @clemfernandez :)

PR #767, opened 2024-08-23
minor mega-linter fixes

readeral:
added trailing slash to staticresources so the regex will target the folder
(although I note that linters running in `project` mode will not use this regex filter, [see here](https://megalinter.io/latest/config-filtering/))

minor adjustments to github workflows to resolve linting issues from shellcheck:
`Double quote to prevent globbing and word splitting`
and from yamllint:
`[document-start] missing document start "---"`


readeral:
I could be wrong on the staticresource, maybe that's a style thing.

nvuillam (approved):

nvuillam:
@readeral many thanks for your PR :)

PR #830, opened 2024-10-13
Remove push on org-monitoring

jawills:
The org monitoring github action does not need to run on every push to the repo. I removed this in the example.

### DIFF
--- a/defaults/monitoring/.github/workflows/org-monitoring.yml
+++ b/defaults/monitoring/.github/workflows/org-monitoring.yml
@@ -18,7 +18,6 @@
 # Doc & support: https://sfdx-hardis.cloudity.com/salesforce-monitoring-home/
 
 on:
-  push:
   # Automatically run every day at midnight
   schedule:
     - cron: "0 0 * * *" # Cron format -> https://crontab.cronhub.io/


defaults/monitoring/.github/workflows/org-monitoring.yml
```diff
@@ -18,7 +18,6 @@
 # Doc & support: https://sfdx-hardis.cloudity.com/salesforce-monitoring-home/
 
 on:
-  push:
```
nvuillam, line 21: The only time when the push is from you is when you configure the job, and it helps to see that the first run works :)
So please can you let the push event here ? :)

defaults/monitoring/.github/workflows/org-monitoring.yml
```diff
@@ -315,7 +314,7 @@ jobs:
         run: |
           echo "Run Monitoring checks against \"$CONFIG_BRANCH\""
           git pull origin ${{ matrix.branch }}
-          sf hardis:th:login
+          sf hardis:auth:login
```
nvuillam, line 317: Good catch :)

nvuillam (changes requested):
Plz check comments :)

defaults/monitoring/.github/workflows/org-monitoring.yml
jawills: This runs when you create the initial branch, as well as when you merge. There is an easy way of running actions with the `workflow_dispatch`. 

If you still want on the push event I can add branch specific events

nvuillam: Hmmmm I forgot that GitHub case is different from other Git providers that allow scheduling from other branches than default one :)

So ok for me to remove push here :)

nvuillam (approved):

PR #837, opened 2024-10-16
initial draft of config helper apex script

ankushsomani09:
This has lwc component which can be used to export and import configuration in text file format having json record data.

###Export###
as of now it export record if given any filter criteria else export all records mentioned in apex class 'objects' variable.
on Succesful execution it download the file of record exported.

###Import###
For import, file which was downloaded from export function, same can be used to import.
It doesn't do diff check, it update/create the config record as given in config downloaed file.


nvuillam:
I'm not sure i understand the goal of this PR, maybe what is already in vscode-sfdx-hardis Data Workbench can handle the same features ?
Plz tell me if this PR needs to be reopen ^^

Issue #902, opened 2024-11-25
`unused-apex-classes` outputs the incorrect CLASSCREATEDON and CLASSCREATEDBY values


k-capehart:
I've double checked that I'm connected to the write org. I ran this command (both via the CLI and the VS Code extension), and while the list of unused classes looks correct, I noticed that the `CreatedDate` and `CreatedBy` values are incorrect.

Here is the result if I query the class.

![Screenshot 2024-11-25 at 1 14 17 PM](https://github.com/user-attachments/assets/ea9bd3ca-54f5-43d1-b5ae-560cbcb56fd5)

Here is how it shows in SFDX Hardis. (I've ommited the Created User's name for privacy but I promise it does not match the above the screenshot). Also, note that the `LATESTJOBDATE` shows correctly as 2024-11-22, but `CLASSCREATEDON` should also be 2024-11-22, but instead shows 2021-09-10.

![image](https://github.com/user-attachments/assets/cc8f8bf8-5e8e-48cc-b23c-971928d1c79f)
![image](https://github.com/user-attachments/assets/57c0ae67-4b63-422d-ae6c-3e981c5c3422)

Also worth mentioning that the same date (2021-09-10) shows for all of the classes, which is not correct. I see at least a couple of instances of (2021-06-28). Normally I would think that is an indication that it's the date of a sandbox refresh, but I triple checked I was connected to Prod, and that date doesn't match any of our sandbox refresh dates anyways.

![Screenshot 2024-11-25 at 1 20 50 PM](https://github.com/user-attachments/assets/ef738022-cd4c-4dd4-a6fa-8759211de633)

nvuillam:
@k-capehart the output created date and created by calculated from MIN(created by of the file in git, created by from the tooling api)

Even with this info, do you still think there is an issue ? :)

k-capehart:
@nvuillam Yeah, the file only has one commit associated with it - 2024-11-20.

![image](https://github.com/user-attachments/assets/f53365ec-216e-4e64-9219-24237f884f31)

k-capehart:
The output of sfdx-hardis does show `(git)` so it must be coming from there somehow. 2021-09-10 matches up nicely with the initial commits for our project, but this class was created only 5 days ago.

Somehow, it is comparing the wrong `gitCreatedDate`

nvuillam:
thank u for the fix :)

k-capehart:
Any time!

PR #903, opened 2024-11-25
check git history for .cls files instead of .cls-meta.xml

k-capehart:
In reference to this Issue: https://github.com/hardisgroupcom/sfdx-hardis/issues/902

Compare `.cls` files when tracking git history instead of `.cls-meta.xml`.

I don't see `findMetaFileFromTypeAndName()` used outside of the `unused-apex-classes` command so I think there is limited impact but let me know if I missed something.


nvuillam:
This method will be used for something else on the future (you will like it^^), bit i'll update your branch to solve the issue you detected :)

Many thanks for your analysis and your PR 😍

nvuillam (approved):

PR #911, opened 2024-12-01
Metadata backup: new option "--full"

nvuillam:
[hardis:org:monitor:backup](https://sfdx-hardis.cloudity.com/hardis/org/monitor/backup/): New mode **--full**, much slower than default filtered one, but that can retrieve ALL metadatas of an org


RaviVaranasi:
I can help with testing this if there is a version available to download

nvuillam:
@RaviVaranasi I tested it on a big org, it should be ok ^^

I release that in a new version in the next 30 mn :)

Issue #924, opened 2024-12-12
JIRA labels for DEPLOY_BEFORE_MERGE = true


0ptaq0:
Some time ago a great feature was introduced to support my exotic way of deployment: SFDX_HARDIS_DEPLOY_BEFORE_MERGE: "true"
When we have a pull/merge request, the merge results (diff) are deployed to the target branch, and if that succeeded the pull/merge request is merged.

I am using that approach but have one ask for additional support for my "deploy before merge" for jira labels.
If `SFDX_HARDIS_DEPLOY_BEFORE_MERGE: "true"` create a tag in a way:
`UPPERCASE(target_branch_name) + "_DEPLOYED"

Currently, we do have: `UPPERCASE(branch_name) + "_DEPLOYED"` which is pointing to the source branch od the merge request and is not really accurate in that case.

nvuillam:
@0ptaq0 that should be hard to implement :)

Time for your first PR to sfdx-hardis ? 😇 

The code building the tag is here -> https://github.com/hardisgroupcom/sfdx-hardis/blob/24b2275b190ad59bc5bb203296fd98c43b5018fb/src/common/ticketProvider/ticketProviderRoot.ts#L27

You can add a condition using [isDeployBefore merge](https://github.com/hardisgroupcom/sfdx-hardis/blob/24b2275b190ad59bc5bb203296fd98c43b5018fb/src/common/gitProvider/index.ts#L192)

To get the targetBranch, you can use 

```typescript
const prInfo = await GitProvider.getPullRequestInfo();
const targetBranch = prInfo?.targetBranch || process.env.FORCE_TARGET_BRANCH
```

0ptaq0:
Hi @nvuillam,

This is my first TS code so I have no idea what I am doing, but I managed to make some changes locally on a feature branch.
I tried to push it but I am getting 403:
```
git push origin features/deploy-before-merge-label
info: please complete authentication in your browser...
remote: Permission to hardisgroupcom/sfdx-hardis.git denied to 0ptaq0.
fatal: unable to access 'https://github.com/hardisgroupcom/sfdx-hardis.git/': The requested URL returned error: 403
```

Can you please give me permission to push?
Or share some other way to have my changes implemented :D

0ptaq0:
Hi @nvuillam, just a reminder, how should I add the PR?

nvuillam:
@0ptaq0 sorry for the delay :)

You need to fork the repo  
[...]

PR #960, opened 2024-12-31
Using target_branch for jira labels when isDeployBeforeMerge flag is true (#924)

0ptaq0:
Related to #924

### DIFF
--- a/src/common/ticketProvider/ticketProviderRoot.ts
+++ b/src/common/ticketProvider/ticketProviderRoot.ts
@@ -2,6 +2,7 @@ import { SfError } from "@salesforce/core";
 import c from "chalk";
 import { Ticket } from "./index.js";
 import { getCurrentGitBranch, uxLog } from "../utils/index.js";
+import { GitProvider } from "../gitProvider/index.js";
 
 export abstract class TicketProviderRoot {
   public isActive = false;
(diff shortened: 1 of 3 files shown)


src/common/ticketProvider/ticketProviderRoot.ts
```diff
@@ -25,9 +26,20 @@ export abstract class TicketProviderRoot {
...
+      const prInfo = await GitProvider.getPullRequestInfo();
+      const targetBranch = prInfo?.targetBranch || process.env.FORCE_TARGET_BRANCH;
+      if (targetBranch) {
+        tag = targetBranch.toUpperCase() + "_DEPLOYED";
+      }
+    }
+
     if (process.env?.DEPLOYED_TAG_TEMPLATE && !(process.env?.DEPLOYED_TAG_TEMPLATE || "").includes("$(")) {
-      tag = process.env?.DEPLOYED_TAG_TEMPLATE.replace("{BRANCH}", currentGitBranch.toUpperCase());
+      const branchToUse = tag.replace("_DEPLOYED", "");
```
nvuillam, line 39: @0ptaq0  Why replacing "_DEPLOYED" ? 
If you calculated and sent your own tag, it should already  contain the string you want , no ? :)

nvuillam (changes requested):
Just a little change (or a good answer to my question) , + add a line in CHANGELOG.md , and we're good to go :)

src/common/ticketProvider/ticketProviderRoot.ts
0ptaq0: Before I added my part we already knew that the branch name is always `currentGitBranch.toUpperCase()`
But when I added the if statement it can be above mentioned or `targetBranch.toUpperCase()`

So in the end when I want to replace `{BRANCH}` I need to know with what I should replace it. That's the only reason why I am replacing _DEPLOYED, to get final branch name.

The other way I thought would be to override `currentGitBranch` with targetBranch in the added if statement. But I didn't want to change that part, as I am not sure of the whole sfdx-hardis structure.

0ptaq0: Hi @nvuillam, is the above okay or should I change it, like overriding or making some different changes?

nvuillam (approved):

nvuillam:
@0ptaq0 i will release a new beta during this evening, and the official release should arrive this week-end :)

0ptaq0:
Nice! Thank you! :)

PR #986, opened 2025-01-10
feat: transform code coverage jsons into cobertura reports

mcarvin8:
Per https://github.com/orgs/hardisgroupcom/discussions/983, adding my apex-code-coverage-transformer plugin to your newer `generateApexCoverageOutputFile` function.

My plugin needs to be ran in the git folder, preferably root folder, to parse the `sfdx-project.json` file for package directories, thus why I'm importing the get repo root function.

My apex-code-coverage-transformer plugin can create 4 possible output formats at the moment: sonar (for sonarqube), cobertura, clover, lcovonly (`.info` file)

Cobertura is suported by most major platforms like GitHub, GitLab, Jenkins, etc. so I'm putting that one in here, but I'm open to suggestion if you want to add more formats to my plugin, use a different format, etc. I'm currently looking into updating my plugin to allow multiple output formats when running a single command, but that's not in the latest build yet.

Let me know what you think @nvuillam and if this is what you meant for the updates to hardis. 

I kinda went on how you integrated sfdx-git-delta into hardis, but let me know if I did anything wrong. First time!

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,21 +5,21 @@ FROM python:3.12.8-alpine3.20
 LABEL maintainer="Nicolas VUILLAMY <nicolas.vuillamy@cloudity.com>"
 
 RUN apk add --update --no-cache \
-            coreutils \
-            git \
-            bash \
-            nodejs \
-            npm \
-            # Required for docker
-            docker \
-            openrc \
-            # Required for puppeteer
-            chromium \
-            nss \
-            freetype \
-            harfbuzz \
-            ca-certificates \
-            ttf-freefont
+    coreutils \
+    git \
+    bash \
+    nodejs \
+    npm \
+    # Required for docker
+    docker \
+    openrc \
+    # Required for puppeteer
+    chromium \
+    nss \
+    freetype \
+    harfbuzz \
+    ca-certificates \
+    ttf-freefont
 
 # Start docker daemon in case mermaid-cli image is used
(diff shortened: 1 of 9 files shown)


nvuillam:
@mcarvin8 sorry I might not have been clear to explain my idea 😢 

I would prefer to avoid to make jobs heavier by adding an extra plugin by default for an optional feature

I thought more about something like : 

- as you can handle hooks, you could create a post-hook in your plugin that would detect if the command start with `hardis:project` or `hardis:org:test:apex` and if there is a file `hardis-report/apex-test-coverage.json`  , then call what you need to call :)

If sfdx-hardis users want to use your feature, they'll just have to install your plugin before calling sfdx-hardis commands :)

Then you can document that on your repo, and i'll make a link to it from sfdx-hardis doc :)

mcarvin8:
Thanks @nvuillam for clearing that up. That's why I was a little confused too.

I already have a hook config in my repo here: [postrun.ts](https://github.com/mcarvin8/apex-code-coverage-transformer/blob/main/src/hooks/postrun.ts)

This runs after the standard sf commands assuming the users create a custom YAML file named `.apexcodecovtransformer.config.json`

I'll need to think of the best way to update the hook to run after the hardis commands so it does not conflict with the standard sf commands, i.e. I don't want the hook to run twice since sfdx-hardis calls the sf-commands.

I'll close this PR here but I'll work on updating my plugin's hook file to use hardis.

PR #1047, opened 2025-01-30
new command to retrieve custom fields usage

Mehdi-Cloudity:
Sample generated report
image


nvuillam:
@Mehdi-Cloudity amazing ! :)

Please can you make the following updates:
- Remove columns FieldId and DependencyId, i don't think anyone will use them  
- Remove PublisherId column: not relevant

After the updates, make a new screenshot, store it in docs/assets/images , then display it in the command documentation
(example: https://github.com/hardisgroupcom/sfdx-hardis/blob/53f8fad05a6ae95bed389a51c10ae52a873c2458/src/commands/hardis/org/diagnose/legacyapi.ts#L31 ) 

Thanks :)

nvuillam (changes requested):
Plz check comment :)

nvuillam (approved):

PR #1079, opened 2025-02-17
Feat: adds --pdf flag for docs commands

matheus-delazeri:
Hi! Firstly, thanks a lot for this tool! :smiley: 

I've created a new command called `flow-full-docs` (not sure about the naming...) that allows the creation of multiple Flows documentations at once, by letting the user select more than one Flow. It also generates the Flow documentation in PDF, using the [md-to-pdf](https://classic.yarnpkg.com/en/package/md-to-pdf), alongside the Markdown version.

I have also implemented it in the **VS Code extension** repository: https://github.com/hardisgroupcom/vscode-sfdx-hardis/compare/main...matheus-delazeri:vscode-sfdx-hardis:main


nvuillam:
@matheus-delazeri  Damn ,that the kind of Pull Request I like ! ;)

If you want the PDF generation to work, I suspect you need to make sure that Flow diagrams are generated as images and not as mermaid code (coz i don't think md-to-pdf will convert it for us) , did you check ?

matheus-delazeri:
Yes! For that, I'm forcing the usage of the Mermaid CLI to generate the Markdown first, because it converts the Mermaid code to a SVG, as stated here: https://github.com/mermaid-js/mermaid-cli?tab=readme-ov-file#transform-a-markdown-file-with-mermaid-diagrams
> This command transforms a markdown file itself. The mermaid-cli will find the mermaid diagrams, create SVG files from them and refer to those in the markdown output.

With that, the `md-to-pdf` package can correctly generate the PDF, as the Mermaid code is replaced by a reference to the SVG in the Markdown file.

So, at the end of the execution of this new command, 3 files are generated:
1. The Markdown file;
2. An SVG file containing the Flow's graph;
3. The PDF file, generated through the Markdown file using `md-to-pdf`.

matheus-delazeri:
@nvuillam I've made the changes you asked!

As for the rest of the documentation, I agree with you! It would be good to have PDFs output for all of them! I've started with Flows because of a customer's demand for PDF documentations of it. But I would be glad to make other PRs to allow it for the other docs :smiley:

nvuillam:
@matheus-delazeri sorry for the delay

I thought about your PR and analyzed it again, and I think that this command is doubling with [those ](https://sfdx-hardis.cloudity.com/hardis/doc/flow2markdown/), and i think that it will make an extra command to maintain whereas we can avoid it ^^

Please can you update the following commands to add a **--pdf** option to have the same result ?
- https://sfdx-hardis.cloudity.com/hardis/doc/flow2markdown/ (you can also update it so it allows to select multiple flows)
- https://sfdx-hardis.cloudity.com/hardis/doc/project2markdown/

I know this is extra work but I'm convinced the result is worth it, and I'll of course highlight your great work to the sfdx-hardis community :)

Does this plan works for you ? :)

matheus-delazeri:
@nvuillam now that you've mentioned, I'm not sure why I haven't done this way from the start haha

I've added the **--pdf** flag for this commands:
- `flow2markdown`
- `project2markdown`: it will generate PDF files for each Apex, Flow, Page and/or Object  markdowns. You would like me to generate for other files also? Like the `index.md` file that is generated.

It added some delay, as the conversion to PDF is not instantaneous, but I think it's worthy! 

About the VS Code extension: you think it would be better to have a different command for running this commands with the flag, like: **Project Documentation (With PDF)**

Or it should be like a configuration for the existing commands?

nvuillam:
I agree with you, 2 extra commands for pdf in sfdx-hardis menu is a good idea :)
I'll test ur PR, merge and release before the weekend :)

matheus-delazeri:
That's great news! I will make the necessary adjustments in the VS Code extension in this time :)

Issue #1093, opened 2025-02-26
Deployment skipped when there are only destructiveChanges filled.


0ptaq0:
We are using the following flag: `SFDX_HARDIS_DEPLOY_BEFORE_MERGE: "true"` and always create a delta package using git delta plugin. (We are not using built-in sfdx hardis functionality to create delta packages as we need to perform some checks on the package before and some other )

After the package is created we move XMLs to the manifest folder:
```
mkdir -p manifest
cp dx-delta/package/package.xml manifest/
cp dx-delta/destructiveChanges/destructiveChanges.xml manifest/
```

For deployment, we are using this command `sf hardis:project:deploy:smart`

It works pretty well, but I believe I spotted a corner case when the destructiveChanges are filled, but package is empty.
Then I am getting:
```
[sfdx-hardis] Empty package.xml: nothing to deploy
[sfdx-hardis] No deployment to perform
[sfdx-hardis] hardis:project:deploy:smart execution time 0:00:04.094
```

Can we also check for destructiveChanges before skipping deployment?

nvuillam:
@0ptaq0 I'll be glad to accept your PR to handle such use case :)

0ptaq0:
Hahaha fair enough, I will try to fix it, but I cannot promise anything :D

0ptaq0:
@nvuillam PR created :)
https://github.com/hardisgroupcom/sfdx-hardis/pull/1098

Tested locally and should be good, but let's see.

0ptaq0:
Changes introduced in main branch and deployed 👌🤝

PR #1098, opened 2025-02-28
handle case when only destructiveChanges are present (#1093)

0ptaq0:
Fix for a corner case when destructive changes are present and package.xml is empty.

Related issue #1093

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ Note: Can be used with `sfdx plugins:install sfdx-hardis@beta` and docker image 
 
 - [hardis:org:test:apex](https://sfdx-hardis.cloudity.com/hardis/org/test/apex/) Take in account `--target-org` option
 - [hardis:org:diagnose:audittrail](https://sfdx-hardis.cloudity.com/hardis/org/diagnose/audittrail/) Fix **monitoringAllowedSectionsActions**
+- Support edge-case when package.xml is empty but destructive changes are present. (see [Github issue](https://github.com/hardisgroupcom/sfdx-hardis/issues/1093))
 
 ## [5.21.0] 2025-02-27
 
--- a/src/common/utils/deployUtils.ts
+++ b/src/common/utils/deployUtils.ts
@@ -166,11 +166,57 @@ export async function smartDeploy(
 ): Promise<any> {
   elapseStart('all deployments');
   let quickDeploy = false;
+
+  // Check package.xml emptiness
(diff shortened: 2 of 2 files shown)


nvuillam:
@0ptaq0 is it tested, and are you confident that it will not trigger regressions ? :)

( I don't really have the time to test myself, and a bug in this method could have awful impacts on all sfdx-hardis projects :D )

0ptaq0:
I tested it only for scenarios like:
- package.xml filled, destructiveChanges.xml empty
- package.xml filled, destructiveChanges.xml filled
- package.xml empty, destructiveChanges.xml empty
- package.xml empty, destructiveChanges.xml filled

But only locally and only in my config. I wouldn't trust myself as this is touching critical functionality, so it would be better if someone could also test it. It can wait, as we know about this, and can live knowing how not to cause it :)

nvuillam:
hmmm i'll generate an alpha that u can use, then after a while i'll make an official release :)

0ptaq0:
Sounds like a plan, thank you! :)

nvuillam:
@0ptaq0 alpha is ready for you to play with :)

https://www.npmjs.com/package/sfdx-hardis/v/5.21.3-alpha202503011721.0

0ptaq0:
Thanks @nvuillam!
I have installed it locally as well as in our project, but it somehow behaves like an old version, meaning the bug still exists.
But when I was checking the version from my Pull Request (local build) it was fixed.

The alpha version is correctly recognized, so I truly don't know what is the issue :<
image

nvuillam:
I'll check if the good code version is in this version, that's really strange :/

0ptaq0:
Thanks! Yes, it is working as expected now. I ran my above-mentioned testing scenarios and they were all good :)

I enabled it for a project. I will let you know if we find any bugs.

package.json
```diff
@@ -44,6 +44,7 @@
     "js-yaml": "^4.1.0",
     "make-fetch-happen": "^14.0.3",
     "markdown-toc": "^1.2.0",
+    "md-to-pdf": "^5.2.4",
```
nvuillam, line 47: why the doubling md-to-pdf ? plz can u fix ? @0ptaq0

0ptaq0:
I have fixed PR, so it contains only one commit with change.
We have not seen any issue with the sfdx-hardis while using the alpha version.

nvuillam:
@0ptaq0 still all good with alpha version ? :)

PR #1108, opened 2025-03-04
ALPHA: handle case when only destructiveChanges are present (#1093)

0ptaq0:
Creating this PR to support alpha branch with my changes


nvuillam:
@0ptaq0 i released a new alpha, you can check :)

PR #1110, opened 2025-03-07
blaissaoui : add Attachment Export files feature

blaissaoui:
New feature : add attachment Export capability to Export files button
first, when configuring export.json file. we will have a choice to select ContentDocument OR Attachment
Next, a new parameter "loadAttachements" will appear in export.json file : 
 - loadAttachements = true  => the script will load only attachment
 - loadAttachements = false => the script will load only ContentDocument
After that, using the exportFile button, the script will query all attachements linked to parentObject (using RestAPI instead of bulkAPI due to a limitation).
Retreived Attachements can be uploaded to salesforce using the same button ImportFiles. But attachement will be created as contentDocument in target Org.

Brahim


nvuillam:
@blaissaoui a million thanks for the PR :)
I'll check asap ^^

nvuillam:
@laissaouibrahim i've not checked in detail yet: do you confirm that if we want to export contentVersion + attachments we need a single export.json file and it can be performed in a single run ?

laissaouibrahim:
@nvuillam : no we should choose at the beggening witch type of document to export (contentdocuement or attachment). After that we will have a dedicated export.json for each Object. 
I prefer to have separated folder because we didn't use the same methods (bulkAPi vs RestAPI) due to limitation on Attachment.
But if you consider that is interesting to retreive both object on one call I can check ;)

image

nvuillam:
@blaissaoui i think this would be best to have all in one, would it be complicated to implement ?

laissaouibrahim:
@nvuillam : ok i'll check and keep you informed :)

PR #1118, opened 2025-03-11
fix: delta after merge is not working as expected

DidZ0:
Fixes an issue where delta deployment after merging a PR isn't calculated correctly.

The issues seems to be the orientation of the delta calculation from "HEAD" to "HEAD^", generating destructive changes if the metadata is new.

inverting the references fixes the issue.

> Thus, the way sfdx-git-delta is used while deploying, it will only generate the delta with the latest commit.
> A temporary workaround would be to activate Squash Commits in the merge request.

### DIFF
--- a/src/commands/hardis/project/deploy/smart.ts
+++ b/src/commands/hardis/project/deploy/smart.ts
@@ -383,8 +383,8 @@ If testlevel=RunRepositoryTests, can contain a regular expression to keep only c
       this.delta = true;
       this.smartDeployOptions.delta = true;
       // Define delta deployment depending on context
-      let fromCommit = 'HEAD';
-      let toCommit = 'HEAD^';
+      let fromCommit = 'HEAD^';
+      let toCommit = 'HEAD';
       if (this.checkOnly) {
         // In deployment check context
         const prInfo = await GitProvider.getPullRequestInfo();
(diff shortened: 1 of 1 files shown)


DidZ0:
@nvuillam 
As far as I'm concerned this isn't even git related, it has more to do with the way sfdx-git-delta is working. 
I'm pretty sure this fix is valid as we indeed want to find what is new from the last commit, not what is deleted / changed if we go back to the last commit.

CF https://github.com/scolladon/sfdx-git-delta?tab=readme-ov-file

> From the project repo folder, the CI pipeline will run the following command:
>
> `sf sgd source delta --to "HEAD" --from "HEAD~1" --output-dir .`
>
> which means:
> `Analyze the difference between HEAD (latest commit) and HEAD~1 (previous commit), and output the result in the current folder.`

This is what you're doing in the check-deploy since you're comparing from the first commit ID of the PR to the last one

nvuillam:
@DidZ0 I generated an alpha version, please can you test with it and confirm that all is ok ? :)

Thanks :)

DidZ0:
@nvuillam So far so good ! It is working as intended, the delta is the same pre and post deployment.
I know this is not a common use case but it would be great to have it.

nvuillam:
@DidZ0 thanks for your feedback, you'll have it very soon, just the time i'm on a laptop 🥳

Issue #920, opened 2024-12-10
Export Files Throws Exception


SegalDaniel:
Hi,

I'm using the export files feature for exporting Account's attachments for my org, but receiving an Unhandled 'error' event in runtime.

[sfdx-hardis] Processing parent records chunk #1 on 12 (1000 records) ...
 throw er; // Unhandled 'error' event
      ^

Error [ERR_STREAM_WRITE_AFTER_END]: write after end
    at _write (node:internal/streams/writable:489:11)
    at Writable.write (node:internal/streams/writable:510:10)
    at DuplexifiedStream.ondata (node:internal/streams/readable:1009:22)
    at DuplexifiedStream.emit (node:events:518:28)
    at addChunk (node:internal/streams/readable:561:12)
    at readableAddChunkPushByteMode (node:internal/streams/readable:512:3)
    at Readable.push (node:internal/streams/readable:392:5)
    at DuplexifiedStream._readStream (/Users/.local/share/sf/node_modules/@jsforce/jsforce-node/lib/util/stream.js:81:18)
    at PassThrough.<anonymous> (/Users/.local/share/sf/node_modules/@jsforce/jsforce-node/lib/util/stream.js:64:18)
    at PassThrough.emit (node:events:518:28)
    at emitReadable_ (node:internal/streams/readable:834:12)
    at process.processTicksAndRejections (node:internal/process/task_queues:89:21)
Emitted 'error' event on PassThrough instance at:
    at PassThrough.onerror (node:internal/streams/readable:1028:14)
    at PassThrough.emit (node:events:518:28)
    at emitErrorNT (node:internal/streams/destroy:170:8)
    at emitErrorCloseNT (node:internal/streams/destroy:129:3)
    at process.processTicksAndRejections (node:internal/process/task_queues:90:21) {
  code: 'ERR_STREAM_WRITE_AFTER_END'

In addition to that, when trying to perform this on a larger amount of records, I got a CSV parsing error, without being able to detect the problematic record/records.

Bulk Query error: Error: Invalid Closing Quote: got "I" at line 8051 instead of delimiter, record delimiter, trimable character (if activated) or comment.

Any ideas? :)

nvuillam:
Hmmm that's strange, did you try several times ?
THe bulk query error could make think about a connection issue ^^

I'll try to reproduce on one of my orgs :)

SegalDaniel:
Yes I have tried it quite a few times and still occurs.
[...]

PR #1119, opened 2025-03-13
Add Attachment Export files feature + improve querying parent record

laissaouibrahim:
1. New feature : add attachment Export capability to Export files button

Using only one clic on export files button, the process is retreiving first Attachment using Rest API ( batchsize =200). After that it will retreive contentDocument using BulkV2 API ( batchsize=1000).

2. the second improvment is about retreiving a large number of ParentRecord. 

When we retreive more than 100000 parentRecord. we got an error [https://github.com/hardisgroupcom/sfdx-hardis/issues/920]. I have added a batchsize=100000 for retreiving ParentRecord. I have tested with 500000 as seen below : 

image


nvuillam:
Thanks a lot @laissaouibrahim :)
Released 🥳

PR #1197, opened 2025-05-04
Feature: add LWC documentation + LWC AI description

tahabasri:
- LWC doc generation is relying on [jsdoc-to-markdown](https://github.com/jsdoc2md/jsdoc-to-markdown)

- LWC doc is getting added under the Code section
image

- Main LWC page is structured like below:
image

- Preview of LWC documentation using jsdoc-to-markdown:
image

- Preview of LWC documentation using AI (with Agentforce):
image

### DIFF
--- a/.github/linters/.cspell.json
+++ b/.github/linters/.cspell.json
@@ -117,6 +117,8 @@
         "LISTVIEWNAME",
         "LVMYAA",
         "Licences",
+        "LWC",
+        "Lwcs",
         "MARKDOWNLINT",
         "MASTERLABEL",
         "MYFOLDER",
--- a/package.json
+++ b/package.json
@@ -42,6 +42,7 @@
     "isomorphic-dompurify": "^2.22.0",
     "jira-client": "^8.2.2",
     "js-yaml": "^4.1.0",
+    "jsdoc-to-markdown": "^9.1.1",
     "make-fetch-happen": "^14.0.3",
     "marked": "^14.1.4",
     "md-to-pdf": "^5.2.4",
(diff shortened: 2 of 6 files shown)


nvuillam (changes requested):
@tahabasri thanks for this great PR :)

Please can you address the following points before I can merge ? :)

- it's a good idea to use jsdoc-to-markdown, but what happens if there is no jsdoc in the code ?
- In the LWC doc page, please can you start with AI generated description ?
- I think that we should send all the LWC to the prompt, in variables LWC_HTML_CODE, LWC_JS_CODE, LWC_JS_META, so we can have an even better description than just with the JS code

tahabasri:
Thanks @nvuillam for reviewing this! Please check my feedback regarding the points mentioned:

1. it's a good idea to use jsdoc-to-markdown, but what happens if there is no jsdoc in the code ?
> For now, if there are no jsdoc compliant comments, the doc file will show "No JSDoc documentation available for this component.". Unfortunately, the only plugin that does [type interference](https://esdoc.org/manual/feature.html#type-inference) is ESDoc, but it's not frequently maintainable and shows lot of security issues when you install it. Current alternative is to rely on AI generated documentation. Will keep looking for other alternatives...

2. In the LWC doc page, please can you start with AI generated description ?
> That's already the case, AI description first, JSDoc generated documentation second (under Summary).
image

3. I think that we should send all the LWC to the prompt, in variables LWC_HTML_CODE, LWC_JS_CODE, LWC_JS_META, so we can have an even better description than just with the JS code
> Last commit should fix that ✅

nvuillam (approved):
Amazing work @tahabasri , thanks a lot for your contribution ! :)

Issue #1205, opened 2025-05-08
Document generation fails on Assignment Rule <criteriaItems> without a <value> element


jcconnell:
Not sure if this issue is specific to our org but wanted to create an issue in case it helps others. Our Monitoring repo started failing roughly two days ago on the generation of docs for Assignment Rules. 

### Command
* `sf hardis:doc:project2markdown`

### Error
```bash
[sfdx-hardis] Generating Approval Processes documentation... (if you don't want it, define GENERATE_AUTOMATIONS_DOC=false in your environment variables)
[sfdx-hardis] Generating Assignment Rules documentation... (if you don't want it, define GENERATE_AUTOMATIONS_DOC=false in your environment variables)
Error (10): Cannot read properties of undefined (reading 'replaceAll')
```

It's failing on the `replaceAll` in the following code block on [line 106](https://github.com/hardisgroupcom/sfdx-hardis/blob/f7acd0b69bd50cea8c9ed75b621024368c02dda7/src/common/utils/rulesBuilderUtil.ts#L106).

```js
formatCriteriaItem(ci: any): string {
    return '(**' + ci.field.split('.')[0] + '**: ' + ci.field.substring(ci.field.indexOf('.') + 1) + ' _' + ci.operation + '_ ' + ci.value.replaceAll(",", ", ") + ')<br>';
  }
```

### Thoughts / Resolution

Our Assignment Rule `< criteriaItems>` were previously defined like:

```xml
<criteriaItems>
    <field>Lead.assignTo__c</field>
    <operation>equals</operation>
</criteriaItems>
```

Changing the definition to the following resolved the issue:

```xml
<criteriaItems>
    <field>Lead.assignTo__c</field>
    <operation>equals</operation>
    <value></value>
</criteriaItems>
```

I'm not the author of the above assignment rule, but I _think_ the intent was to compare the field to a `null` value. It looks like the change to `rulesBuilderUtil.ts` expects a `value` element to be present, even if it's indicating a null/empty-value.

nvuillam:
@jcconnell meanwhile you can define the following CI/CD variable as workaround, to not generate such doc

`GENERATE_AUTOMATIONS_DOC=false`

mpyvo:
Yes, I’m on it!

nvuillam:
@jcconnell released in latest patch :)

Please can you confirm it's now ok for you use case ? :)

PR #1207, opened 2025-05-09
Fix: fixed no value in criteria item

mpyvo:
Fixed issue with empty value for the criteria item:

Знімок екрана 2025-05-09 о 12 33 34

Fixes https://github.com/hardisgroupcom/sfdx-hardis/issues/1205


nvuillam (approved):
thanks for the fix :)

PR #1214, opened 2025-05-10
Feature: langchain integration to support multiple AI providers

anushpoudel:
- Added langchain provider support + some key standard model parameters like temperature, max tokens, etc.
- Implemented for ollama, openai, and anthropic providers for now - extending to other providers should be fairly straightforward with the factory pattern implementation as most providers support the same standard parameters as listed here: https://js.langchain.com/docs/concepts/chat_models. Happy to add more extensions when I have some time on my hand.
- Updated docs for langchain support

@nvuillam I've left the previous openai implementation as is - didn't want to remove it and have things break for existing implementations that use the openai provider directly.

FYI: Did test it briefly with for ollama and openai key since I had one - haven't tested it for anthropic.


nvuillam:
@anushpoudel this is a piece of art, thank you, ccan't wait to try it  ! :)

Please can you fix the MegaLinter issue, this is cspell complaining about words he doesn't know

To fix that, get the updated .cspell.json in MegaLinter artifacts -> https://github.com/hardisgroupcom/sfdx-hardis/actions/runs/14944148510/artifacts/3098831188

Overwrite the one in your branch, check with git diff if these are really new words, or if there are typos, then commit - push :)

nvuillam (approved):
AMAZING :)

nvuillam:
@anushpoudel i will release it later today or tonight, when I'll have completed my current task allowing to override locally the prompts used to generate AI doc :)

PR #1216, opened 2025-05-10
Bugfix: use explicit type casting to avoid deep type analysis

anushpoudel:


nvuillam (approved):
thanks for the fix :)

PR #1248, opened 2025-05-26
Optimize member sorting during manifest generation

derroman:
Previously, when generating the org manifest, the members list for metadata types (e.g., ListView, CustomLabel) was sorted upon each individual member addition within the main processing loop. This caused redundant sorting operations, significantly degrading performance—especially in Salesforce orgs with a large number of components. In one recent real-world scenario, this step alone took over 60 minutes to complete.

This commit refactors the logic in buildOrgManifest to:

1. Defer sorting (sortCrossPlatform) until after all elements for a metadata type have been collected, thus performing sorting only once per type.

With these changes, the sorting process now completes in just a few moments, drastically improving the manifest generation speed and efficiency.

### DIFF
--- a/src/common/utils/deployUtils.ts
+++ b/src/common/utils/deployUtils.ts
@@ -1127,7 +1127,7 @@ export async function buildOrgManifest(
         // Add member in existing types
         const members = matchTypes[0].members || [];
         members.push(element.fullName);
-        matchTypes[0].members = sortCrossPlatform(members);
+        matchTypes[0].members = members;
         parsedPackageXml.Package.types = parsedPackageXml.Package.types.map((type) =>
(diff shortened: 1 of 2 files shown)


src/common/utils/deployUtils.ts
```diff
@@ -1140,7 +1140,13 @@ export async function buildOrgManifest(
         parsedPackageXml.Package.types.push(newType);
       }
     }
-
+    // Sort members for each type once after the loop
+    for (const type of parsedPackageXml.Package.types) {
+      if (type.members && Array.isArray(type.members)) {
```
nvuillam, line 1145: if you want to even improve perfs, you can sort only the types that are in mdTypes :)

nvuillam (approved):
Well played @derroman :)

PR #1276, opened 2025-06-18
Apply bypass to VR/Triggers

Mehdi-Cloudity:


nvuillam (changes requested):
Nice work @Mehdi-Cloudity :)
Please can you apply the following updates ?

- Prompt user if he wants to retrieve metadatas (on a CI/CD project they are already here, so it will improve perfs to not retrieve them)
- display "generated by sfdx-hardis" or "updated by sfdx-hardis" depending on what the command did
- Do not update a source file in case the diff is just formatting 
  - Example: image
- When updating a VR, is it possible to make a carriage return after the added ByPass check condition, to improve readability ?
  - image
- Please be careful to not break triggers
  - Example here with amalto stuff removed: image
- Add "PAD" check to detected if there is already a ByPass

Mehdi-Cloudity:
> Nice work @Mehdi-Cloudity :) Please can you apply the following updates ?
> 
> * Prompt user if he wants to retrieve metadatas (on a CI/CD project they are already here, so it will improve perfs to not retrieve them)
>   
>   * If users says no, you can find metadatas using method `const fileMetadata = await MetadataUtils.findMetaFileFromTypeAndName("ApexClass", cls.Name);`
> * display "generated by sfdx-hardis" or "updated by sfdx-hardis" depending on what the command did
> * Do not update a source file in case the diff is just formatting
>   
>   * Example: image
> * When updating a VR, is it possible to make a carriage return after the added ByPass check condition, to improve readability ?
>   
>   * image
> * Please be careful to not break triggers
>   
>   * Example here with amalto stuff removed: image
> * Add "PAD" check to detected if there is already a ByPass
> 
> Note: i updated your branch with a few updates ^^

1. Optional retrieve : implemented
2. Change action in credit text: implemented
3. Do not update a source file in case the diff is just formatting : no code update if the bypass isn't implemented, this must be the result of the retrieve
4. Formatting post credit: implemented
5. Please be careful to not break triggers: I was unable to replicate this, from the screenshot the bypass wasn't implemented because PAD has a method with keyword "bypass", this must be the result of the retrieve, can you please recheck ?
6. Verify PAD: implemented based on "PAD.can", the keyword bypass is already in place

nvuillam (commented):
thanks @Mehdi-Cloudity :)

PR #1308, opened 2025-07-10
Restore connected apps after refresh SB

zzyviolette:


zzyviolette:
TODO : Use --metadata-dir for deployment, since --manifest doesn’t work with sandboxes that don’t have source tracking enabled.

PR #1343, opened 2025-08-12
Adds option to hide apex source code in generated documentation

kg345:
Adds a `--hide-apex-code` flag to `sf hardis:doc:project2markdown` command.

This would allow a user to control whether apex source code is outputted into the apex code documentation pages.


nvuillam:
@kg345 thanks for the PR :)

Please can you also add the capability to disable apex output in doc if there is an env variable **HIDE_APEX_CODE** defined with value "true" ?

it will be useful for users using sfdx-hardis Monitoring to update their doc every night :)

nvuillam:
@kg345 available in beta version, and in a few days in official release :)

PR #1397, opened 2025-09-08
Enable ermaid cli in the Docker images

### LINKED ISSUE #1383: Mermaid flow visualisation not working when use hardis docker image in azure devops pipeline
In my case flow diff images are not being generated since I started to use hardis as docker image in my azure pipeline as:

```
jobs:
  - job: DeploymentCheck
    timeoutInMinutes: 150
    container:
      image: hardisgroupcom/sfdx-hardis-ubuntu:latest
```

If I force MERMAID_MODES=docker, the volume mappig does not work properly. Copilot explaination:

>1. Azure Pipelines Docker volume mount isolation
Azure Pipelines runs in a containerized agent (vsts_azpcontainer), and Docker inside Docker (DinD) can cause volume mounts to fail silently if the path isn't accessible from the Docker daemon's perspective.
> 2. Docker daemon context mismatch
The path /__w/1/s/hardis-report/flow-diff might be valid in the agent container, but not visible to the Docker daemon, which runs in a different context.

```
Use $(System.DefaultWorkingDirectory) instead of hardcoded paths: 
docker run --rm -v "$(System.DefaultWorkingDirectory)/hardis-report/flow-diff:/data"
```

Which **I could reproduce and confirm** in my pipeline experiments. But this is obviously not a feasible solution.

If I force MERMAID_MODES=cli, I got:
>Error: Failed to launch the browser process!
/home/vsts_azpcontainer/.cache/puppeteer/chrome-headless-shell/linux-131.0.6778.204/chrome-headless-shell-linux64/chrome-headless-shell: error while loading shared libraries: libglib-2.0.so.0: cannot open shared object file: No such file or directory

> TROUBLESHOOTING: https://pptr.dev/troubleshooting

    at Interface.onClose (file:///home/vsts_azpcontainer/.npm/_npx/668c188756b835f3/node_modules/@puppeteer/browsers/lib/esm/launch.js:303:24)
    at Interface.emit (node:events:531:35)
    at Interface.close (node:internal/readline/interface:534:10)
    at Socket.onend (node:internal/readline/interface:260:10)
    at Socket.emit (node:events:531:35)
    at endReadableNT (node:internal/streams/readable:1698:12)
    at process.processTicksAndRejections (node:internal/process/task_queues:90:21)

But I'm not sure if the supporting of cli mode would be a good idea inside of hardis container.

So for now I don't have any good solution for that.

step307:
Resolves #1383

### DIFF
--- a/Dockerfile-ubuntu
+++ b/Dockerfile-ubuntu
@@ -8,16 +8,38 @@ ENV SF_DATA_DIR=/usr/local/lib
 
 # Update package list and install required packages
 RUN apt-get update && apt-get install -y --no-install-recommends \
-            git \
-            curl \
-            gnupg \
-            # Required for Python
-            python3 \
-            python3-pip \
-            python3-venv \
-            # Required for docker
-            docker.io \
-            fonts-liberation && \
+    git \
+    curl \
+    gnupg \
+    # Required for Python
+    python3 \
+    python3-pip \
+    python3-venv \
+    # Required for docker
+    docker.io \
+    fonts-liberation \
+    # Required for mermaid cli
+    libglib2.0-0 \
+    libnss3 \
+    libx11-6 \
+    libx11-xcb1 \
+    libxcb1 \
+    libxcomposite1 \
+    libxcursor1 \
+    libxdamage1 \
+    libxext6 \
+    libxfixes3 \
+    libxi6 \
+    libxtst6 \
+    libdrm2 \
+    libgbm1 \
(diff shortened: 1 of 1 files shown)


nvuillam:
@step307 thanks :)

please can you also add mermaidjs install in the Dockerfile , with `npm install @mermaid-js/mermaid-cli --global` , to improve runtime performances ? :)

step307:
@nvuillam, done

Issue #1092, opened 2025-02-26
ORG_LIMITS throwing error instead of warning


gustavozbulhoes:
The `hardis:org:monitor:limits` explained [here](https://sfdx-hardis.cloudity.com/hardis/org/monitor/limits/) should throw error only if limit is > 100% according to [this article](https://sfdx-hardis.cloudity.com/salesforce-monitoring-org-limits/). 
For the org that I'm testing it the DataStorageMB is in 91.56% but that is throwing error. See screenshot from Github Action:

![Image](https://github.com/user-attachments/assets/80163c86-08ca-4637-a89d-af4bb83510cf)

From my tests I also realized that for error in org limit the Github Action throws an error. I would love to not throw an error in this case (sometimes the limit error can be a know issue that will not be fixed soon). So would be really nice to have one of the options below:
1. Variable to define if ORG_LIMITS should throw an error in the whole action if some limit is overreached. In my case for example I do not wanna see errors in the Monitoring Github Actions everyday (because in Github seems the monitoring failed but it did not) because DataStorageMB limit is overreached.
2. Similar to `monitoringExcludeUsernames` for AUDIT_TRAIL, a property to skip specified ORG_LIMITS would be really helpfull

Appreciate your help.

gustavozbulhoes:
Yes @nvuillam that works.

marcosboger:
@nvuillam could I pick this up to work on? 

Seems like a good one for first-time contribution, not so complex yet meaningful

nvuillam:
@marcosboger please be my guest :)

nvuillam:
@marcosboger please also handle it from .sfdx-hardis.yml config file, so we'll be able to make it configurable in the Extension UI :)
( jsonschema needs to be updated ^^ )

marcosboger:
Sounds good. I'll draft something and create a PR early to see if in the right direction

nvuillam:
@marcosboger don't hesitate to write me if you have any question :)

nvuillam:
Fixed in #1422 :)

PR #1422, opened 2025-09-25
Fix: adjust LIMIT_THRESHOLD_ERROR env var + align docs threshold values

marcosboger:
Relates to https://github.com/hardisgroupcom/sfdx-hardis/issues/1092

As part of working on the issue, I noticed that var `LIMIT_THRESHOLD_ERROR` was not actually being used and `LIMIT_THRESHOLD_WARNING` was duplicated. Also, the docs on [salesforce-monitoring-org-limits](https://sfdx-hardis.cloudity.com/salesforce-monitoring-org-limits/) had different threshold values than those defined in [limits](https://sfdx-hardis.cloudity.com/hardis/org/monitor/limits/) and in code.

This PR addresses those two points. I'll create a separate one for the configurability discussed in the issue itself.

### DIFF
--- a/package.json
+++ b/package.json
@@ -76,7 +76,9 @@
   "resolutions": {
     "@langchain/core": "0.3.57",
     "form-data": "4.0.4",
-    "tmp": "0.2.5"
+    "tmp": "0.2.5",
+    "prebuild-install/tar-fs": "2.1.4",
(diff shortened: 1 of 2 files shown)


nvuillam (approved):
Great, thanks for the fix :)

PR #1442, opened 2025-10-06
Feature: Automate SF decomposed metadata command (beta version)

smo17:
Automating sf beta decomposed metadata command with sfdx-hardis command:
`sf hardis:project:metadata:activate-decomposed`

https://developer.salesforce.com/docs/atlas.en-us.sfdx_dev.meta/sfdx_dev/sfdx_dev_ws_decomposed_md_types.htm#sfdx_dev_ws_start_decompose

The hardis command starts by checking local salesforce project for already decomposed metadata from the current beta list:
CustomLabels, PermissionSet, ExternalServiceRegistration, SharingRules, Workflow

If one or more found, it will list the already decomposed metadata and ignore those.
If none or partial found, it will prompt the user once to confirm operation on eligible metadata then proceed with decomposing the applicable metadata(s).
Depending on source data, this operation make take some time and a status is provided in logs/UI.


smo17:
> @smo17 Looks amazing :)
> 
> I see you directly calls WebSocketClient, but you don't need to ^^
> 
> * You don't need to create new WebSocketClient, it is done automatically from [init hooks](https://github.com/hardisgroupcom/sfdx-hardis/blob/main/src/hooks/init/start-ws-client.ts)
> * When you want to start a new section, just call uxLog("action",...): every next uxLog entries ("log", "success", "warning") will be embedded in the same section, until there is a new section opened by another uxLog("action") or a prompts"
> * When you want to prompt something to the user, just call await prompts(...)
> 
> Only reason you could have to directly call WebSocketClient is to send report file paths to the UI using `WebSocketClient.sendReportFileMessage`
> 
> Please can you apply the described changes ?

@nvuillam Good catch. The changes are now done, thanks to review again

nvuillam:
@smo17 here is my result for the first run  :/

image

Would it be possible to:
- Allow to select the metadatas types that we want to decompose ? (use `await prompts({type: "multi-selelect" ... )`
- Provide more details when there are errors ?
- To execute commands, maybe you could call directly `execCommand()` method ? (start with `echo|y` if you need confirmation, or maybe try with --json )

Thanks :)

smo17:
@nvuillam Oh that's weird 😕
Tests are successful on a dev and sandbox org, see sample result for a sandbox:
image

Can you share your sfdx-project.json ? I suspect maybe the error is occurring because of a different package directory path name.
Can you also share the logs ? normally errors will always appear in the logs below and I will make sure to have them in the UI screen too.

I will add the metadata type selection and check the execute commands
Thanks

nvuillam:
@smo17 here is my sfdx-project.json, very classic... but i also have a second package, that's what might cause the crash ^^

```json
{
  "packageDirectories": [
    {
      "path": "force-app",
      "default": true
    },
    {
      "path": "survey-force"
    }
  ],
  "name": "hardisgroup",
  "namespace": "",
  "sfdcLoginUrl": "https://login.salesforce.com",
  "sourceApiVersion": "63.0"
}
```

For your SF Cli command, you can list the sfdx-project packages using this code: `const packageDirs = this.project?.getPackageDirectories();`

Not: requires your command class to have `  public static requiresProject = true;`

smo17:
@nvuillam PR updated with these changes:
- Calling `execCommand()` directly
- Use `getPackageDirectories()` to read project folders
- improved error handling/logging
- Allow to multiselect metadata to decompose with selects all as default

image

Could not reproduce the exit code 1 error but normally these changes should give more details in the logs.
Try out the command again and let me know

Thanks

nvuillam:
@smo17 i pushed a commit to handle the case where default org is source-tracked :)

Please can you validate my updates ? :)

image

smo17:
@nvuillam Ah I see it now, the source tracking was causing the error code 1, thanks for catching this.
The new updates are all good 😊

PR #1447, opened 2025-10-09
Cast descriptions to string before formatting

### LINKED ISSUE #1446: BUG: project2markdown crashes when some field description is numeric
If sime field deascription conrains only numbers project2markdown crashes with error "Error (10): str.replace is not a function"

Root cause:
Numeric fiel description is casted to a numebr in js and the when passed to [`mdTableCell`](https://github.com/hardisgroupcom/sfdx-hardis/blob/8835c95469fa5a37d14694415222ec764428fee1/src/common/gitProvider/utilsMarkdown.ts#L84) crashes on `return str.replace(/\n/gm, "<br/>").replace(/\|/gm, "");`

step307:
Resolves #1446


nvuillam:
Who the hell describe their fields with just numbers ? 🤣

PR #1495, opened 2025-10-26
Fix Validation Rules formula field in doc generation

matheus-delazeri:
Hey! This is a minor fix to the Validation Rules doc generation to prevent it from breaking when a formula contains line breaks.

**Original VR**

image

**Before**

image

**After**

image


nvuillam (approved):
Well played @matheus-delazeri ,thanks a lot for your PR :)

PR #1525, opened 2025-11-10
Fix File Export SOQL with LIMIT Clause

mvcmania:
Handles the given SOQL with LIMIT clause, replaces LIMIT with given limit and finally put the LIMIT clause at the end of the SOQL
Issue is,when you specify the limit clause in the file export soql, plugin is also adding another limit clause and it clashes and throwing MALFORMED QUERY exception like below : 
Screenshot 2025-11-10 at 11 04 29

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 Note: Can be used with `sfdx plugins:install sfdx-hardis@beta` and docker image `hardisgroupcom/sfdx-hardis@beta`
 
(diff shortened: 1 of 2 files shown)


mvcmania:
@nvuillam you might be right in general, but one example that i came across is , during org migration.We sometime have to provide sample files (not big enough, yet not small number of files) to  be confirmed and approved by stakeholder. For those kind of particular case, limit could be very useful.
Regarding to multispace normalization, i came across that on bulkquery method which appends the ORDER BY. 
Thank you again for this plugin. Im glad if i made some useful contribution to the project.

nvuillam:
@mcarvin8 I agree, the use case is real, thanks a lot for your contribution :) 

It will be released in the next patch ^^ (today or tomorrow :) )

Issue #1538, opened 2025-11-15
bulkQueryByChunks generates invalid SOQL with duplicate WHERE clauses during pagination


prabhumadhan:
When exporting files using sf hardis:org:files:export with a SOQL query that already contains a WHERE clause, the pagination logic create invalid SOQL queries with duplicate WHERE clauses instead of using AND to combine conditions.

**Steps to Reproduce:**

  1. Create an export configuration with a SOQL query containing a WHERE clause:
  ```
{
    "sfdxHardisLabel": "CaseFilesExport",
    "soqlQuery": "SELECT Id, CaseNumber, ContactId, AccountId FROM Case WHERE Status = 'Closed'",
    "fileTypes": "all",
    "outputFolderNameField": "AccountId"
  }
```

  2. Run the export command:
  `sf hardis:org:files:export`

  3. Monitor the logs when processing more than 100,000 records (triggering pagination)

**Expected Behavior:**

  Pagination queries should correctly combine WHERE clauses using AND:
  `SELECT ... FROM Case WHERE Status = 'Closed' AND Id > '5009000001P56RSAAZ' ORDER BY Id LIMIT 100000`

**Actual Behavior:**

  The query is generated with duplicate WHERE clauses, causing a SOQL syntax error:
  `SELECT ... FROM Case WHERE Status = 'Closed' WHERE Id > '5009000001P56RSAAZ' ORDER BY Id LIMIT 100000`

**Root Cause:**

  In lib/common/utils/apiUtils.js, the bulkQueryByChunks function at line 122 always appends WHERE Id > '...' without checking if the original
  query already contains a WHERE clause:

  ```
if (lastRecordId) {
      soqlQueryWithLimit = `${soqlQuery} WHERE Id > '${lastRecordId}' ORDER BY Id`;
  }
```

**Suggested Fix:**

Check if the query already has a WHERE clause and use AND instead:

  ```
if (lastRecordId) {
      // Check if query already has a WHERE clause to use AND instead of WHERE
      const hasWhere = soqlQuery.toUpperCase().includes('WHERE');
      const connector = hasWhere ? 'AND' : 'WHERE';
      soqlQueryWithLimit = `${soqlQuery} ${connector} Id > '${lastRecordId}' ORDER BY Id`;
  }
```

**Environment:**
  - sfdx-hardis version: 6.12.2
  - Salesforce API version: 65.0
  - Platform: macOS (Darwin 24.6.0)

**Impact:**

[...]

PR #1539, opened 2025-11-16
Fix: Resolve duplicate WHERE clause in bulkQueryByChunks pagination

prabhumadhan:
## Summary

Fixes issue #1538 - Resolves duplicate WHERE clause problem in `bulkQueryByChunks` function

## Changes

- Check if SOQL query already contains WHERE clause before pagination
- Use AND connector for pagination when WHERE exists
- Use WHERE connector when query has no WHERE clause

## Problem

The `bulkQueryByChunks` function in `src/common/utils/apiUtils.ts` was always appending `WHERE Id > '...'` for pagination, causing invalid SOQL syntax when the original query already contained a WHERE clause.

**Before:**
```typescript
if (lastRecordId) {
  soqlQueryWithLimit = `${soqlQuery} WHERE Id > '${lastRecordId}' ORDER BY Id`;
}
```

**After:**
```typescript
if (lastRecordId) {
  // Check if query already has a WHERE clause to use AND instead of WHERE
  const hasWhere = soqlQuery.toUpperCase().includes('WHERE');
  const connector = hasWhere ? 'AND' : 'WHERE';
  soqlQueryWithLimit = `${soqlQuery} ${connector} Id > '${lastRecordId}' ORDER BY Id`;
}
```

## Impact

- Fixes invalid SOQL syntax when paginating filtered queries
- Resolves issue with file exports exceeding 100k records with filters
- Maintains backward compatibility for queries without WHERE clauses

## Related Issue

https://github.com/hardisgroupcom/sfdx-hardis/issues/1538

### DIFF
--- a/src/common/utils/apiUtils.ts
+++ b/src/common/utils/apiUtils.ts
@@ -148,7 +148,7 @@ export async function bulkQueryByChunks(
     let soqlQueryWithLimit = `${soqlQuery} ORDER BY Id`;
     if (lastRecordId) {
       // Check if query already has a WHERE clause to use AND instead of WHERE
-      const hasWhere = soqlQuery.toUpperCase().includes('WHERE');
+      const hasWhere = soqlQuery.toUpperCase().includes(' WHERE ');
       const connector = hasWhere ? 'AND' : 'WHERE';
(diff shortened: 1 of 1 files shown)


nvuillam:
@prabhumadhan will be included in the next release today or tomorrow :)

prabhumadhan:
thank you @nvuillam :)

PR #1588, opened 2025-12-11
Fix auth to Jira Data Center

nvuillam:
- Auth to jira data center 
- Instructions to test a development version of sfdx-hardis from a CI/CD pipeline


nvuillam:
Finally working :)

Issue #1517, opened 2025-11-07
New possible Feature : Field Usage


MathieuRodriguesCloudity:
Hello,

Field Usage feature was available in Optimizer and there is an option in SF Inspector by object.

In search of a new way to get Field Usage for all objects, I found this article : https://www.pablogonzalez.io/how-to-calculate-field-utilization/

Can be an inspiration to add this feature in the plugin ? :)

Fernando-Fernandez:
@MathieuRodriguesCloudity  Isn't that feature already implemented here?  https://sfdx-hardis.cloudity.com/hardis/doc/fieldusage/

nvuillam:
@Fernando-Fernandez this one is about dependencies, i think @MathieuRodriguesCloudity is interested about which fields are used for storage :)

Fernando-Fernandez:
Hi @nvuillam , I've submitted this PR implementing the feature as described:  https://github.com/hardisgroupcom/sfdx-hardis/pull/1627

MathieuRodriguesCloudity:
Hi @Fernando-Fernandez, good news :) 
I am eager to use this new feature !

Fernando-Fernandez:
@MathieuRodriguesCloudity I already loving this new feature! :)
It should become a nice replacement for that old package FieldTrip that is no longer available.

datamike:
@nvuillam It looks like this Issue should be closed?

nvuillam:
@datamike indeed, @Fernando-Fernandez did a great job ;)

PR #1627, opened 2025-12-29
Implemented Obj field usage

Fernando-Fernandez:
Implemented new Object Field Usage feature:  this will count how many records has nulls in each field and report the percentage of nulls.
If run with the parameter --field, it will count distinct values in the field (up to 1000) and report the percentages of each value.
The goal is to help determine if a field is being used or not and whether some values dominate (possibly indicating that users are either entering dummy data or perhaps that the field is not that useful).

https://github.com/hardisgroupcom/sfdx-hardis/issues/1517

### DIFF
--- a/package.json
+++ b/package.json
@@ -76,6 +76,7 @@
   "resolutions": {
     "@langchain/core": "0.3.80",
     "form-data": "4.0.5",
+    "qs": "6.14.1",
     "tmp": "0.2.5",
     "prebuild-install/tar-fs": "2.1.4",
     "@puppeteer/browsers/tar-fs": "3.1.1",
--- a/yarn.lock
+++ b/yarn.lock
@@ -10799,10 +10799,10 @@ puppeteer@>=8.0.0:
     puppeteer-core "24.14.0"
     typed-query-selector "^2.12.0"
 
-qs@^6.10.1, qs@^6.10.3:
-  version "6.14.0"
-  resolved "https://registry.yarnpkg.com/qs/-/qs-6.14.0.tgz#c63fa40680d2c5c941412a0e899c89af60c0a930"
-  integrity sha512-YWWTjgABSKcvs/nWBi9PycY/JiPJqOD4JA6o9Sej2AtvSGarXxKC3OQSk4pAarbdQlKAh5D4FCQkJNkW+GAn3w==
+qs@6.14.1, qs@^6.10.1, qs@^6.10.3:
+  version "6.14.1"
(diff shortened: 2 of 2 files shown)


src/commands/hardis/doc/object-field-usage.ts
```diff
@@ -0,0 +1,393 @@
...
+
+  public static examples = [
+    '$ sf hardis:doc:object-field-usage --sObject Account',
+    '$ sf hardis:doc:object-field-usage --target-org myOrgAlias --sObject CustomObject__c',
+    '$ sf hardis:doc:object-field-usage --sObject Account --field SalesRegionAcct__c',
+  ];
+
+  public static flags: any = {
+    'target-org': requiredOrgFlagWithDeprecations,
+    sObject: Flags.string({
```
nvuillam: plz can u use 's-objects' or 'objects' ? :)
It would be nice to be able to call the command for multiple objects in one call

In case the option is not sent, you can also prompt the user to select objets that he/she wants to analyze :)

See example of objects selection here: https://github.com/hardisgroupcom/sfdx-hardis/blob/e346bfcd7444c6730950d8af9e2313b0604e6f05/src/commands/hardis/org/diagnose/storage-stats.ts#L202

nvuillam (changes requested):
Looks amazing @Fernando-Fernandez , thanks :)

Please can you handle the remarks I added in the review ? :)

Also, if before running the queries, it would be possible to estimate the number of API calls that will be consumed and ask for user confirmation ?

Fernando-Fernandez:
@nvuillam I believe I implemented all the requested improvements.

- Changed the flag from --sObject to --Objects. 
- It now works for multiple objects in one call.
- Before running the queries, the estimated number of API calls that will be consumed is shown and it prompts the user for confirmation.
- It generates one CSV per object and a summary CSV then aggregates all of them in a single XLS file.
- Renamed the --field parameter to --fields and made it accept multiple fields.
- When no field is specified, it will run it for all relevant fields.
- When --objects is not specified, it will prompt the user to select objects.
- When --objects is specified with no value, like above, it will prompt the user to select objects.

Fernando-Fernandez:
> @Fernando-Fernandez just tested it, it seems to work well :)
> 
> last update; in th XSLX, please can you put the summary tab in the first position ?

Done. ✅

nvuillam:
I make a last enhancement then i'll merge :)

Fernando-Fernandez:
> Wonderful PR @Fernando-Fernandez :) i'll release it in beta until the next minor version :)

Awesome! Great ideas, by the way!

PR #1653, opened 2026-01-13
Feature/include original translations

JulianoMedeirosMachado:
Hi!

This enhancement allows retrieving the original Custom Label value, which is normally missing from the generated translation files. For example, if a label is defined with the value "Example" under the en_US locale, that value is not included in the respective en_US translation file. With this change, the original translation file will now include the label value, resulting in more standardized and consistent output.

As an example, before the translation XML returned the following for the default language in en_US.translation-meta.xml:
```xml
<customLabels>
<label/>
<name>ExampleLabel</name>
</customLabels>
```

After this feature a new XML will be generated using `CustomLabels.labels-meta.xml`, `original.translation-meta.xml`:
```xml
<customLabels>
<name>ExampleLabel</name>
<label>Example Text</label>
<shortDescription>Example Description</shortDescription>
</customLabels>
```

Just let me know if anything needs tweaking!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,8 @@
 
 Note: Can be used with `sfdx plugins:install sfdx-hardis@beta` and docker image `hardisgroupcom/sfdx-hardis@beta`
 
+## [6.22.0] 2025-01-13
+
 - [hardis:project:configure:auth](https://sfdx-hardis.cloudity.com/hardis/project/configure/auth/):
   - Use Connected App by default because issues have been reported with External Client Apps in some Salesforce editions / setups.
   - Add copy tags around values so VsCode SFDX-Hardis can display a copy button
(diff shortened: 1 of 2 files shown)


nvuillam:
@matheus-delazeri what do you think ? (i don't use this command that you created so you know better ^^ )
If it adds a new behavior, is it ok to have it by default, or do you think it's better to activate it with an additional CLI option ?

matheus-delazeri:
@nvuillam yes, that's something I've being struggling with but didn't have the time to implement. Thanks a lot @JulianoMedeirosMachado!! 

I do think it can be active by default, as the main usage for this function is to get all the translations of the custom labels and we will not lose any info by that. In my opinion, is ready to merge :)

nvuillam:
thanks for your contribution @JulianoMedeirosMachado ,I'll release it in the next patch :)

PR #1687, opened 2026-01-30
bugfix: OS agnostic folderscan for local items in clean manageditems

SalikLP:
The local folder scan was not working on windows, so the sf hardis project clean manageditems deleted localitems also.
This should be an OS agnostic fix where the cwd input to glob is the folder to check instead of process.cwd(). Then the fullpath is found using path.join().


nvuillam:
@SalikLP thanks for the PR :)

You can test with sfdx-hardis@beta , it will be released later in the week-end :)

PR #1701, opened 2026-02-05
JSON.Stringify() object to avoid textInit.startsWith error in uxLog

SalikLP:
The uxLog throws an textInit.startsWith error when trying to parse an object


nvuillam (approved):
Thanks for the fix :)

Issue #1604, opened 2025-12-19
hardis:doc:project2markdown linter errors leading to failure in .md generation for apex classes and triggers


maksym-petrov-ct:
After launching `sf hardis:doc:project2markdown` there are log lines like this:
```
Using temp directory /tmp/sfdx-hardis-ml4hv
Error generating Apex documentation: [
  "Error(s) occurred while parsing files. Please review the following issues:",
  "Source file: force-app/main/default/classes/my_class.cls\nException: Error parsing Apex Body. Line 378:52 - missing ';' at '('\n"
]
```
First of all, I have to tell  that all class files and triggers were just retrieved from the org and are deployable. 
The line 378 in question is nothing suspicious:
```
    private static void AdjustSL_SE_Date_FromJanuary(...
```
Offset :52  points to 'r' in word January above.

**It ends up not generating any apex or trigger .md files. Objects, flows, etc are generated.**

If the my_class.cls is deleted along with it's -meta.xml, then another set of errors for triggers appears, like this:
```
"Source file: force-app/main/default/triggers/Account_trigger_vod.trigger\nException: Error parsing Apex Body. Line 1:39 - extraneous input 'bulk' expecting '('\n",
```

Here is the 1st line of example trigger:
```
trigger Account_trigger_vod on Account bulk (before delete, after delete) {
```
I tried to delete all triggers mentioned in error message. It leads to apex .md files are generated, **but triggers .md are still not generated,** despite no visible errors.
The generation time for `sf hardis:doc:project2markdown` is always 6-7 minutes in my project (no AI integration) and is unrelated to errors above. 

```
node --version
v24.11.1

sf --version
@salesforce/cli/2.114.5 wsl-x64 node-v24.11.1

sf plugins
@salesforce/commerce 252.0.0
@salesforce/sfdx-scanner 4.2.0 (latest-beta)
code-analyzer 5.0.0 (5.0.0)
community 3.2.32 (3.2.32)
dev 2.5.1 (2.5.1)
packaging 2.23.5
sfdmu 4.38.0
sfdx-git-delta 6.24.0
sfdx-hardis 6.14.2
signups 2.6.43 (2.6.43)
texei-sfdx-plugin 2.8.3
```

nvuillam:
@maksym-petrov-ct > sfdx-hardis relies on https://github.com/cesarParra/apexdocs to generate apex doc, so the parsing problem is to address there

cc @cesarParra

[...]

PR #1718, opened 2026-02-13
Added "Trigger" detection and generation of docs in hardis:doc:project2markdown

maksym-petrov-ct:
Related to https://github.com/hardisgroupcom/sfdx-hardis/issues/1604

1. Added "Trigger" detection in returnApexType() in src/common/utils/projectUtils.ts 
Triggers starting with trigger are now classified as type "Trigger" instead of falling through to the default "Class".
2. src/commands/hardis/doc/project2markdown.ts — Three changes:
a) Copy ApexDocGen's triggers/ output:
@cparra/apexdocs now supports triggers and generates markdown for them,
but it outputs them to a triggers/ subfolder. The code was only copying from miscellaneous/,
so trigger docs were silently lost. Now both miscellaneous/ and triggers/ are copied to the docs/apex/ folder.
b) Include trigger names in apexClassNames for relationship detection:
Previously, trigger names were filtered out of apexClassNames, so they couldn't
appear in cross-reference diagrams. Now both classes and triggers participate in relationship tracking.
c) Fallback for triggers without ApexDocGen output (robustness):
If a trigger somehow doesn't get an .md file from ApexDocGen (e.g., ApexDocGen fails),
a basic markdown doc is still created from scratch with the trigger's source code, mermaid diagram, and AI description placeholder.

### DIFF
--- a/.github/linters/.cspell.json
+++ b/.github/linters/.cspell.json
@@ -355,6 +355,7 @@
     "apexlog",
     "apexp",
     "apexpages",
+    "apexdevtools",
     "apextest",
     "apiversion",
     "apos",


nvuillam:
@maksym-petrov-ct you have MegaLinter issues , please can you fix them ? :)

src/common/utils/projectUtils.ts
```diff
@@ -137,24 +137,42 @@ export async function listPageFiles(packageDirs) {
   return pageFiles.sort();
 }
 
+// npmjs.com/package/@apexdevtools/apex-parser may give more robust result, but if use it, better to start from project2markdown.ts
```
maksym-petrov-ct: should i remove the comment ? megalinter does not like it.

PR #1741, opened 2026-02-26
Added OpenAI Codex as AiProvider

SalikLP:
We are already paying for ChatGPT with Codex in my company so I thought why not use that 🙂 

This PR adds a minimal CodexProvider integration in the same way ass the existing AI Providers.

What’s included
Add Codex SDK dependency.
Add new CodexProvider extending AiProviderRoot.
Register Codex in AI provider selection flow.
Added relevant config/env variables.
Supports both local authentication (codex is already installed and you are logged in via ChatGPT subscription) and API key.

### DIFF
--- a/config/sfdx-hardis.jsonschema.json
+++ b/config/sfdx-hardis.jsonschema.json
@@ -139,6 +139,33 @@
       "title": "LangChain base URL",
       "type": "string"
     },
+    "useCodexDirect": {
+      "$id": "#/properties/useCodexDirect",
+      "description": "Set to true to enable the direct Codex connector (mirrors USE_CODEX_DIRECT env var).",
+      "title": "Enable Codex",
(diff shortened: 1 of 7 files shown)


nvuillam:
@SalikLP as Codex is provided by OpenAI, what will integration with Codex SDK bring more than existing integrations to OpenAi via langchainJs ?

SalikLP:
For CI this will not make a difference, but for local running e.g. for local project2markdown it will reuse your existing codex auth/credentials so there is no need to provide an API key 🙂 My company is paying for ChatGPT Business but not for the OpenAI Platform, so I didn't have an API key, hence I added this option 🙂

nvuillam (approved):
Your reasons are good enough for me , thanks for the PR :)

Issue #1772, opened 2026-03-10
Login to non-branch org failing with empty username


SalikLP:
I'm trying to deploy to a second org (training org) from the main branch.
I have duplicated my Azure deploy pipeline and then manually set the ORG_ALIAS to pick up the correct env variables and key file.

That seems to not be enough so I have also tried setting a mixture of env variables like TARGET_USERNAME, INSTANCE_URL, CONFIG_BRANCH, CI_COMMIT_REF_NAME, but I cant get it to pick up my second org config. 
I tried some debugging and the username here: [authUtils.ts#L168](https://github.com/hardisgroupcom/sfdx-hardis/blame/130aaa8295d0bfdf01cbaa5a858346a3a6d45d7f/src/common/utils/authUtils.ts#L168) is always '' (empty string). 
Is there a bug in the resolution or am I doing something wrong?
The way I read the Username resolution it seem like TARGET_USERNAME is used as a boolean OR condition together with isDevHub and if either is truthy then take the devHubUserName from config (which in my case is blank string) but it never gets to the targetUsername from the config (which in this case would also have to overwritten by CONFIG_BRANCH)

PR #1773, opened 2026-03-10
Ensure that target username is not lost to devhub

SalikLP:
Small proposed fix for https://github.com/hardisgroupcom/sfdx-hardis/issues/1772 where the targetUsername is not overwritten by devHubUsername on non-branch config


nvuillam (approved):
LGTM :)

PR #1779, opened 2026-03-11
Backup external client apps before sandbox-refresh

SalikLP:
Added the metadata needed to backup/restore External Client Apps to the default package-metadatas-to-save.xml.
Also added a small warning to remind users to comment/remove the tags that will cause the deploy to fail. 
It could probably be automated, but this was the simplest solution for now 😄


nvuillam:
@SalikLP I did all that (and even more ^^) in alpha version that will be released in a few days :)

Run before-refresh and after-refresh from alpha version (`sf plugins install sfdx-hardis@alpha`) and you'll see :)

(we do stuff like prompting for secrets for ext client apps to backup, delete connected apps that have been migrated as an External Client app that we want to backup, etc.... :) )

PR #1780, opened 2026-03-11
feat: add org monitoring for underused perm sets

mcarvin8:
Here's my 1st attempt at adding a new monitoring command.

Add a new command to `sf hardis:org:diagnose:underusedpermsets` which runs 4 SOQLs in the org:
1. Return permission sets (custom) which are assigned to 0 users, **excluding any perm set in a perm set group**
2. Return permission sets (custom) which are assigned to 5 or under users, **excluding any perm set in a perm set group** (configurable threshold via a new environment variable PERMSET_LIMITED_USERS_THRESHOLD)
3. Return permission set groups (custom) which are assigned 0 users.
4. Return permission set groups (custom) which are assigned to 5 or under users (configurable threshold via a new environment variable PERMSET_LIMITED_USERS_THRESHOLD)

Multtiple SOQLs are needed as the second one can't get the 0 users assigned perm sets. This is something apart of my current monitoring exporter, so I think this slides in nicely with sfdx-hardis's default org monitoring and promotes permission set maturity in an org.

Testing is SAT on my end while connected to my org, but please verify on your end. I can provide some screenshots, but probably not the Excel files from my work org's set themselves.

Let me know if I missed any required documentation updates or updates for adding it to the All Org Monitoring command.

### DIFF
--- a/docs/salesforce-monitoring-underused-permsets.md
+++ b/docs/salesforce-monitoring-underused-permsets.md
@@ -1,23 +1,29 @@
 ---
 title: Underused Permission Sets (Salesforce monitoring)
-description: Schedule weekly checks for permission sets with few or no users
+description: Schedule weekly checks for permission sets and permission set groups with few or no users with sfdx-hardis Monitoring
 ---
 <!-- markdownlint-disable MD013 -->
 
 ## Detect underused permission sets
 
-Permission sets with zero or few users may be candidates for cleanup or consolidation. This command helps identify technical debt and supports permission set hygiene.
+Permission sets and Permission Set Groups with zero or few assigned users may be candidates for cleanup or consolidation. This command helps identify technical debt and supports permission hygiene.
 
 **What it detects:**
 
-- **0 users:** Permission sets and Permission Set Groups with no assignments
-- **Low usage:** Permission sets and groups assigned to 5 or fewer users (configurable via `PERMSET_LIMITED_USERS_THRESHOLD`)
+- **0 users (error):** Permission sets and Permission Set Groups with no assignments at all
+- **Low usage (warning):** Permission sets and groups assigned to `PERMSET_LIMITED_USERS_THRESHOLD` or fewer users (default: **5**)
 
-**Permission Sets:** Custom only (NamespacePrefix = null, LicenseId = null), not profile-owned, not in groups. PSL-linked and managed package permission sets are excluded.
+**Scope:**
 
-**Permission Set Groups:** Custom only (NamespacePrefix = null). Managed package groups are excluded.
+- **Permission Sets:** Custom only (`NamespacePrefix = null`, `LicenseId = null`), not profile-owned, not included in a Permission Set Group. PSL-linked and managed package permission sets are excluded.
+- **Permission Set Groups:** Custom only (`NamespacePrefix = null`). Managed package groups are excluded.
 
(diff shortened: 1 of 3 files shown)


nvuillam:
@mcarvin8 great, thanks :)

Why not also querying PermSet Groups ?
I'm afraid that it returns false positive, is PS A has 0 assignments, but PSG X contains PS A and has 200 user assignments

nvuillam:
and also maybe add a way to ignore some perm sets ? (those who are from managed packages, or those who have been created especially for a service user)

I would think:
- automatically exclude managed PS
- allow some variable UNDERUSED_PERMISSION_SETS_IGNORE with a custom list of PS to ignore

mcarvin8:
> @mcarvin8 great, thanks :)
> 
> Why not also querying PermSet Groups ? I'm afraid that it returns false positive, is PS A has 0 assignments, but PSG X contains PS A and has 200 user assignments

Both queries contains this logic, so even if they are under-assigned direclty on the PS, if they are included in a permission set group, they will not be included in the results.

```
        where PermissionSetId NOT IN (
            SELECT PermissionSetId
            FROM PermissionSetGroupComponent
        )
```

I can see how to include permission set groups in the calculation, so if a PS group is under-assigned, it should be flagged as well.

mcarvin8:
> and also maybe add a way to ignore some perm sets ? (those who are from managed packages, or those who have been created especially for a service user)
> 
> I would think:
> 
> * automatically exclude managed PS
> * allow some variable UNDERUSED_PERMISSION_SETS_IGNORE with a custom list of PS to ignore

Let me confirm if the exclusing of managed PS is working, as I'm seeing some permission sets in my org get flagged which I don't have access to, so perhaps something's off in the excluding managed PS logic in the under 5 SOQL.

And I can work on some variable to allow users to specifiy ones to ignore.

mcarvin8:
https://github.com/hardisgroupcom/sfdx-hardis/pull/1780/commits/31d45879c613556735de452734063d04c6819420 - This addresses pulling permission sets that have licenses. So no perm sets should be coming from managed packages anymore. Verified against my org that all perm sets found are custom.

https://github.com/hardisgroupcom/sfdx-hardis/pull/1780/commits/2eb8322748b3b986bb5a2417cc36c386cd6155fb - Added 2 queries for permission set groups under the same criteria (0 users or under X threshold)

https://github.com/hardisgroupcom/sfdx-hardis/pull/1780/commits/1eabb62782c4fe5ac9f8558fda92ec77c8bc6271 - adds new environment variable to ignore permission sets. I tested with this variable UNDERUSED_PERMISSION_SETS_IGNORE=Manage_Billing,User_Stitch_Prod

Let me know if there's anything else, but that should cover not querying perm sets in licenses/managed packages, excluding underassigned perm sets if they're in a perm set group, including perm set groups in these calculations,  and allowing the user to ignore specific perm sets/perm set groups from this monitoring

Everything looks good on my end after linking to my local sf cli and running it against my org. CSV contains all perm sets/groups I can see (custom ones) and ignores perm sets in my env variable.

nvuillam:
@mcarvin8  Available in sfdx-hardis@alpha , and soon in v7 :)
Note i made a small update to add a summary section at the end :)

image

mcarvin8:
Look goods to me 🚀

PR #1783, opened 2026-03-13
Removed invalid Env Var already handled in getPullRequestInfo()

SalikLP:
In a deployment scenario the getEnvVar("SYSTEM_PULLREQUEST_PULLREQUESTID") resolves to $(System.PullRequest.PullRequestId) which in line [#540](https://github.com/hardisgroupcom/sfdx-hardis/blob/cb5333cdea1d97c95cd6ff14f59ef13943ab52d4/src/common/gitProvider/azureDevops.ts#L540) gives NaN when trying to interpret as number.

There is already a more solid check in getPullRequestInfo() so this env just overwrites that and cause the message post to fail.

### DIFF
--- a/src/config/index.ts
+++ b/src/config/index.ts
@@ -271,7 +271,9 @@ export async function getReportDirectory() {
 export function getEnvVar(envVarName: string): string | null {
   const varValue = process.env[envVarName] || null;
   // Avoid Azure cases that sends the expression as string if variable not defined
-  if (varValue && varValue.includes(`(${envVarName}`)) {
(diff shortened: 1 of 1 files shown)


nvuillam:
getEnvVar is supposed to not return a value when it contains stuff like that, i'd prefer you fix getEnvVar direcly because such issue could happen somewhere else :)

nvuillam:
For example, check if the value matches $(...) and return null if it's the case

SalikLP:
I have rewritten it to now match on $(...) 🙂

PR #1793, opened 2026-03-16
feat: monitor custom apex api versions in the org

mcarvin8:
Add new org monitor to query all custom Apex classes and Apex triggers running on old API versions (configurable threshold via CLI flag or environment variable).

I know you have a similar [deprecatedapiversions monitor](https://sfdx-hardis.cloudity.com/hardis/org/diagnose/legacyapi/), but that seems to focus on API calls using outdated API versions over custom Apex classes deployed to an org that are still running on older versions. I think these can complement each other, but let me know what you think.

This will just monitor custom Apex classes and triggers in the org running on some configurable threshold, so teams could set a higher threshold to be more proactive with keeping their Apex running on the latest and greatest.

By default, test classes will be excluded but they can be included via the `--includetestclasses`/`-i` Boolean flag.

Side-note: I'm adding a `.gitattributes` file here as I keep getting random files showing as "changed" due to line-endings as I'm on Windows. In my git repos, we just normalize to LF for Unix compatibility since we run Unix containers, so I assume that's safe for this repo and this can help Windows contributors from seeing unrelated files from showing as "changed" due to line-endings.


mcarvin8:
conflicts resolved and env variable renamed to DEPRECATED_APEX_API_VERSION

PR #1804, opened 2026-03-19
Migrated to OpenAI ResponsesAPI 

SalikLP:
## Summary

Migrate the direct OpenAI provider from Chat Completions to the Responses API and add support for configurable `service_tier` and `reasoning` options.

## What Changed

- Switched `OpenAiProvider` to use `openai.responses.create(...)` instead of `chat.completions.create(...)`
- Kept the default OpenAI model as `gpt-4o-mini`
- Added optional support for:
  - `OPENAI_SERVICE_TIER` / `openaiServiceTier`
  - `OPENAI_REASONING_EFFORT` / `openaiReasoningEffort`
- Added schema support for the new config keys in `sfdx-hardis.jsonschema.json`
- Updated OpenAI setup and environment variable documentation to describe the new options

## Implementation Notes

- Supported service tiers: `auto`, `default`, `flex`
- Supported reasoning efforts: `low`, `medium`, `high`
- Invalid or unsupported optional values are ignored rather than breaking provider initialization
- Response parsing now reads from the Responses API output (`output_text`)

## Why

This aligns the OpenAI integration with the current API surface and exposes newer request controls for latency/cost/performance tuning on supported models and projects.
e.g. This allows using the flex tier which is half the cost (but slower)

## Docs

Documentation was updated to cover both environment-variable and `.sfdx-hardis.yml` configuration for the new OpenAI options.

### DIFF
--- a/config/sfdx-hardis.jsonschema.json
+++ b/config/sfdx-hardis.jsonschema.json
@@ -180,6 +180,28 @@
       "type": "string",
       "default": "gpt-4o-mini"
     },
+    "openaiServiceTier": {
+      "$id": "#/properties/openaiServiceTier",
+      "description": "Service tier used for direct OpenAI prompts. Equivalent to OPENAI_SERVICE_TIER env var.",
+      "title": "OpenAI service tier",
+      "type": "string",
+      "enum": [
+        "auto",
+        "default",
+        "flex"
+      ]
+    },
+    "openaiReasoningEffort": {
+      "$id": "#/properties/openaiReasoningEffort",
+      "description": "Reasoning effort used for direct OpenAI prompts on supported models. Equivalent to OPENAI_REASONING_EFFORT env var.",
+      "title": "OpenAI reasoning effort",
+      "type": "string",
+      "enum": [
+        "low",
+        "medium",
+        "high"
+      ]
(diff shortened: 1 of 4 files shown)


nvuillam:
Great :)
Do you confirm backward compatibility with existing configurations? :)

SalikLP:
Yes, I have tried running different combinations:

- Only defining the OPENAI_API_KEY and USE_OPENAI_DIRECT (fallback to gpt-4o-mini in code)
- Above + defining gpt-5-mini model in sfdx-hardis.yml
- Defining env OPENAI_MODEL=gpt-5.4-mini
- OPENAI_MODEL + OPENAI_SERVICE_TIER=flex
- OPENAI_MODEL  + OPENAI_REASONING_EFFORT=medium
- OPENAI_MODEL  + OPENAI_SERVICE_TIER=flex + OPENAI_REASONING_EFFORT=medium
- OPENAI_MODEL=gpt-4.1-mini (non-reasoning model)

But if you use a model that does not support either flex or reasoning while defining one of those ENV you will get an error in the console like: `Error while calling AI provider: 400 Unsupported parameter: 'reasoning.effort' is not supported with this model.`

PR #1805, opened 2026-03-19
feat: monitor org deployments

mcarvin8:
Using the Tooling API, monitor deployments and validations (CheckOnly) in an org and report high-level stats such as totals, success rate %, average deployment time (minutes), and average pending time (minutes).

This allows users to track deployment metrics over time periods like 1 day, 1 week, or the full period (30 days). Salesforce only maintains deployment records for 30 days.

{DDD14994-8D6E-4731-8ACC-1756C75ED119}

## Tooling API Issue (`src/common/utils/apiUtils.ts`)

**What happens**

When using conn.tooling.query() (jsforce / @salesforce/core), Tooling API queries return only 100 records per request by default, even when more records exist.

Salesforce only maintains 30 days worth of deployment records, so users who want to query the full amount cannot due to this limitation. My current org has around 320 deployment records over this past month, which is pretty standard.

**Why it matters**

- The REST API’s default batch size is 2000 records.
- The Tooling API appears to use a smaller default (often 100 records).
- conn.tooling.query() and conn.tooling.queryMore() do not expose a way to set the batch size.
- As a result, queries against Tooling objects (e.g. DeployRequest, ApexClass, ApexLog) can cap at 100 records unless pagination is used correctly.

**Evidence**

- Users with 200+ DeployRequest records only see 100 rows in reports.
- Salesforce REST API docs: [Query Options Header](https://developer.salesforce.com/docs/atlas.en-us.api_rest.meta/api_rest/headers_queryoptions.htm) supports Sforce-Query-Options: batchSize=2000 (range 200–2000).
- Stack Overflow: [Do we have setQueryOptions for tooling API?](https://salesforce.stackexchange.com/questions/218864/do-we-have-setqueryoptions-for-tooling-api) shows the REST Tooling API accepts this header.

**Workaround used**

Instead of conn.tooling.query(), we call conn.request() directly with:

1. Sforce-Query-Options: batchSize=2000 to request up to 2000 records per batch.
2. Manual pagination using nextRecordsUrl and the same header for follow-up requests.

### DIFF
--- a/src/commands/hardis/org/diagnose/apex-api-version.ts
+++ b/src/commands/hardis/org/diagnose/apex-api-version.ts
@@ -6,9 +6,9 @@ import c from 'chalk';
 import { uxLog, uxLogTable } from '../../../../common/utils/index.js';
 import { soqlQueryTooling } from '../../../../common/utils/apiUtils.js';
 import { generateCsvFile, generateReportPath } from '../../../../common/utils/filesUtils.js';
-import { getNotificationButtons, getOrgMarkdown, getSeverityIcon } from '../../../../common/utils/notifUtils.js';
+import { getSeverityIcon } from '../../../../common/utils/notifUtils.js';
 import { NotifProvider, NotifSeverity } from '../../../../common/notifProvider/index.js';
-import { setConnectionVariables } from '../../../../common/utils/orgUtils.js';
+import { prepareOrgNotificationContext } from '../../../../common/utils/orgNotificationContext.js';
 import { CONSTANTS, getEnvVar } from '../../../../config/index.js';
 import { isTestClass } from '../../../../common/utils/apexLimitUtils.js';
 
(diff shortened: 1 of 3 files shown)


nvuillam:
@mcarvin8 please can you make MegaLinter happy ? :)

The issue found is duplicate code, if it's real duplicate you can factorize, otherwise ignore duplicate using `/* jscpd:ignore-start */` and `/* jscpd:ignore-end */` (ask you code agent, he'll know ^^ )

image

mcarvin8:
> @mcarvin8 please can you make MegaLinter happy ? :)
> 
> The issue found is duplicate code, if it's real duplicate you can factorize, otherwise ignore duplicate using `/* jscpd:ignore-start */` and `/* jscpd:ignore-end */` (ask you code agent, he'll know ^^ )
> 
> image

Looks like this is fixed by the latest commit after I checked the MegaLinter job results

PR #1843, opened 2026-04-08
feat: monitor the apex flex queue

mcarvin8:
Add new org monitoring for the [Apex Flex Queue](https://help.salesforce.com/s/articleView?id=platform.code_apex_flex_queue.htm&type=5), which is something my org monitors for so we can create a case when the queue exceeds the max of 100 jobs.

SOQL: 

```
SELECT Id, ApexClassId, ApexClass.Name, JobType, CreatedDate FROM AsyncApexJob WHERE Status = 'Holding' ORDER BY CreatedDate ASC
```

{5B72199F-294B-4053-A53B-27F680823C61}


nvuillam (approved):
Good PR, thanks ! :)
Will be in the next release :)

PR #1865, opened 2026-04-24
Bugfix/empty page prefix and storage format conversion errors

SalikLP:
Two related fixes to sf hardis:doc:mkdocs-to-confluence so that publishing works cleanly when the environment is configured for a shared Confluence space.

#### 1. Empty CONFLUENCE_PAGE_PREFIX / CONFLUENCE_PAGE_SUFFIX are now respected

`||` treated an empty string as "not set" and silently fell back to the [Doc]  default, which made it impossible to disable the prefix. Switched to `??` so an explicitly empty value is kept as-is and only undefined triggers the default. Same change applied to the suffix.

#### 2. Confluence Storage Format conversion hardened

Several markdown patterns produced XHTML that Confluence rejected with `Content contains unsupported extensions and cannot be edited in Fabric editor` (HTTP 400 Bad Request). Fixed:

- Inline code spans are now emitted via placeholders and XML-escaped, so bodies like `List<LogRequest>` no longer leak raw angle brackets, and later italic/bold passes can't splice * or _ inside a code span.
- Bold and italic regexes now forbid newlines inside the span — prevents a stray ** or * from greedily matching across unrelated paragraphs/list items.
- XHTML sanitization pass added (sanitizeConfluenceXhtml): preserves CDATA, escapes bare &, self-closes void elements (`<br>, <hr>, `), and escapes any `<...>` chunk that isn't a recognized storage-format / `ac:` / `ri:` tag.
- Table cells run through the same tag allow-list so Apex signatures like `List<LogRequest>` inside a table don't get parsed as unknown markup.

### DIFF
--- a/src/commands/hardis/doc/mkdocs-to-confluence.ts
+++ b/src/commands/hardis/doc/mkdocs-to-confluence.ts
@@ -232,8 +232,8 @@ The command orchestrates interactions with MkDocs configuration, Markdown conver
     this.confluenceClientId = getEnvVar('CONFLUENCE_CLIENT_ID') || null;
     this.confluenceClientSecret = getEnvVar('CONFLUENCE_CLIENT_SECRET') || null;
     this.confluenceParentPageId = (flags.confluenceParentPageId as string) || getLocalizedEnvVar('CONFLUENCE_PARENT_PAGE_ID') || null;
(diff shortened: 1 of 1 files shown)


nvuillam (approved):
Thank u very much for the fixes 😊😊

PR #1892, opened 2026-05-06
Keep local metadata when on managed object

SalikLP:
Enhances hardis:project:clean:manageditems to preserve custom items defined on managed objects: 
- compound-filename metadata (Layouts, QuickActions) where the object is namespaced but the item isn't,
- Workflow / SharingRules XMLs that contain at least one non-namespaced sub-item. 
 
Adds a generic config-table mechanism so new metadata types are one-line additions. Also a tiny log-message fix in emptyitems.ts.


nvuillam:
Damn, i was sure i already took it in account, thanks 🤗

nvuillam (changes requested):
Please can u ask you agent to create test cases covering this command, including your latest updates ? :)

SalikLP:
Yes of course 👍

nvuillam (approved):
Great :)

PR #1903, opened 2026-05-11
fix: expand minimizeProfile node removal and fix off-by-one bug

SalikLP:
- Extended the list of Profile XML sections stripped by minimizeProfile to cover four additional nodes: agentAccesses, customPermissions, flowAccesses, and ServicePresenceStatusAccesses - all of which can be managed via Permission Sets and should not live in minimized profiles.
- Fixed an off-by-one bug where a profile with exactly one removed section was not written to disk (removed.length > 1 should have been removed.length >= 1), causing silent no-ops when only a single node type was present.
- Added a unit test suite (test/common/utils/profileUtils.test.ts) covering all four new nodes, the off-by-one fix, pre-existing nodes, the no-change case, and multi-section removal.


nvuillam:
@SalikLP i have many updates in a sub branch, that should fix most of the MegaLinter issues, i'll merge your PR then :)

nvuillam (approved):

PR #1934, opened 2026-06-03
Fixed the hardcoded suffixes to query actual names from the org

SalikLP:
It seems that the ECA "Satellite" metadata suffix is hardcoded while in practice it varies a lot. so the backup of the metadata does not always work, but the console shows errors like:
`Entity of type 'ExtlClntAppConfigurablePolicies' named 'sfdx_hardis_uat_defPlcy' cannot be found`
`Entity of type 'ExtlClntAppGlobalOauthSettings' named 'sfdx_hardis_uat_defGlblOauthSet' cannot be found`
Where the real name is `sfdx_hardis_uat_defaultPolicy` and `sfdx_hardis_uatGlblOAuth` respectively. 
In this PR we query the org for the actual names and match with the "parent" ECA name.


nvuillam:
@SalikLP thanks :)
Do you confirm it has been tested ? :)

SalikLP:
I tried it on my laptop where it worked without issues. But I haven't had other people test it 😊

nvuillam (approved):

PR #1944, opened 2026-06-11
fix: handle directories in hiddenitems clean

### LINKED ISSUE #1943: hardis:project:clean:hiddenitems fails with EISDIR when glob matches a directory
## Bug description

`sf hardis:project:clean:hiddenitems` can fail with `EISDIR: illegal operation on a directory, read` when a Salesforce DX project contains a directory whose name ends with one of the scanned file extensions.

The command scans this pattern:

```text
**/*.{app,cmp,evt,tokens,html,css,js,xml}
```

It then calls `fs.readFile(...)` on each match. Static resource dependencies can contain directories such as:

```text
force-app/main/default/staticresources/<static-resource>/bower_components/history.js
```

That path is a directory, but it matches the `.js` suffix. The command then tries to read the directory as a file and exits with `EISDIR`.

## Steps to reproduce

1. In a Salesforce DX project, add or retrieve a static resource folder containing a directory ending in `.js`, for example `force-app/main/default/staticresources/example/bower_components/history.js/`.
2. Run:

```sh
sf hardis:project:clean:hiddenitems
```

## Actual behavior

The command fails:

```text
[sfdx-hardis] Removing hidden dx managed source files
Error (EISDIR): EISDIR: illegal operation on a directory, read
```

## Expected behavior

The command should ignore directories matched by the glob and only read regular files before checking whether their content starts with `(hidden)`.

## Suggested fix

Use `nodir: true` in the `glob(...)` call and keep a defensive `stat.isFile()` guard before `fs.readFile(...)`, because component folder cleanup can remove files after the glob result has already been computed.

## Validation idea

Add a regression test that creates:

- one directory whose name ends with a scanned extension, such as `FolderWithXmlExtension.xml/`
- one real file whose content is `(hidden)`
- one real non-hidden file

The command should ignore the directory, remove the hidden file, and keep the non-hidden file.

manoelcalixto:
## Summary

- make `hardis:project:clean:hiddenitems` ignore glob matches that are directories
- add a regression test for directories whose names end with scanned extensions

## Root cause

`hardis:project:clean:hiddenitems` scans `**/*.{app,cmp,evt,tokens,html,css,js,xml}` and then calls `fs.readFile(...)` on every match. Static resource dependencies can contain directories such as `bower_components/history.js`, which match the `.js` suffix but are not files. Reading that path throws `EISDIR: illegal operation on a directory, read`.

## Fix

The command now asks `glob` to return files only with `nodir: true`, and also keeps a defensive `stat.isFile()` guard before reading each match.

Fixes #1943

## Validation

- `yarn mocha "test/commands/hardis/project/clean/hiddenitems.test.ts"`
- `yarn eslint "src/commands/hardis/project/clean/hiddenitems.ts" "test/commands/hardis/project/clean/hiddenitems.test.ts"`
- `yarn tsc -p . --pretty false --incremental false`
- `yarn tsc -p ./test --pretty false --noEmit`

### DIFF
--- a/src/commands/hardis/project/clean/hiddenitems.ts
+++ b/src/commands/hardis/project/clean/hiddenitems.ts
@@ -126,11 +126,15 @@ In agent mode, all interactive prompts are skipped and default values are used.
     /* jscpd:ignore-end */
     const rootFolder = path.resolve(this.folder);
     const findManagedPattern = `**/*.{${HIDDEN_SOURCE_FILE_EXTENSIONS}}`;
-    const matchingCustomFiles = await glob(findManagedPattern, { cwd: rootFolder, ignore: GLOB_IGNORE_PATTERNS });
+    const matchingCustomFiles = await glob(findManagedPattern, { cwd: rootFolder, ignore: GLOB_IGNORE_PATTERNS, nodir: true });
     let counter = 0;
     for (const matchingCustomFile of matchingCustomFiles) {
       const matchingCustomFilePath = path.join(rootFolder, matchingCustomFile);
       if (!fs.existsSync(matchingCustomFilePath)) {
+        continue;
+      }
+      const fileStats = await fs.stat(matchingCustomFilePath);
+      if (!fileStats.isFile()) {
         continue;
       }
       const fileContent = await fs.readFile(matchingCustomFilePath, 'utf8');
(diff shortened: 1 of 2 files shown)


nvuillam:
@manoelcalixto please can you merge conflicts ? :)

manoelcalixto:
> @manoelcalixto please can you merge conflicts ? :)

Sure, I’ll resolve the conflicts and update the PR

manoelcalixto:
> @manoelcalixto please can you merge conflicts ? :)

Conflicts resolved

nvuillam (approved):
thanks a lot for the fix :)

PR #1957, opened 2026-06-17
Harden Flow visualizer labels for non-string XML values

manoelcalixto:
## Summary

This hardens Flow visual documentation generation when a Flow label is parsed from XML as a non-string value.

The Flow visualizer previously assumed labels were always strings and called string replacement methods directly. Some XML values, such as `<label>false</label>`, can be parsed as booleans. In that case, Mermaid label sanitization can fail even though the Flow metadata is otherwise valid.

## Changes

- Add `normalizeMermaidLabel()` to convert unknown label values into stable display strings before Mermaid sanitization.
- Support primitive values, arrays, and common Flow value-object shapes such as `stringValue`, `elementReference`, `targetReference`, `name`, and `#text`.
- Keep the existing Mermaid escaping behavior after normalization.
- Add unit coverage for boolean Flow labels.

## Validation

- `node --import tsx/esm --no-warnings=ExperimentalWarning ./node_modules/mocha/bin/mocha.js test/common/utils/flowParser.test.ts`
- `npx eslint src/common/utils/flowVisualiser/flowParser.ts test/common/utils/flowParser.test.ts`


manoelcalixto:
@nvuillam Hi, from what I've seen, the problems you encountered with CI aren't related to the changes, is that correct?

nvuillam:
it should be ok now , i fixed the CVEs ^^

nvuillam (approved):
thanks for the PR :)

Issue #1946, opened 2026-06-12
Make hardis compatible with git worktrees


SalikLP:
I would like to be able to use hardis in different git worktrees. When I run the sf hardis work new (Click: New User Story) then it fails with the error: 
```
git checkout main
Command error (50s)
fatal: 'main' is already used by worktree at 'C:/gitrepos/salesforce'
```

nvuillam:
@SalikLP that's a great idea, but i don't have a lot of worktree experience, would you like to implement the PR ? (i think it should also work with hardis:work:save)

PR #1968, opened 2026-06-23
Support git worktrees by branching from origin/<target>

SalikLP:
## What

`hardis:work:new` now creates the new work branch from `origin/<target>` instead of
checking out the target branch locally first. This makes it work when the target branch
is already checked out in another git worktree, which git otherwise refuses ("` fatal: '<target>' is already used by worktree at <path>`").

## Why

When you use git worktrees, the same branch cannot be checked out in two worktrees at
once. The old flow (`gitCheckOutRemote(target)` -> `gitPull()` -> `ensureGitBranch(new)`)
checked out the target branch in the current worktree, so starting a new User Story
failed whenever the target branch was open elsewhere.

## How

- Replace the checkout + pull + create-branch sequence in `work/new.ts` with a single new
  helper `createWorkBranchFromTarget(branchName, targetBranch)`.
- `createWorkBranchFromTarget` (in `common/utils/index.ts`):
  - Refuses early with a clear message if the **work** branch is already checked out in
    another worktree.
  - Fetches `origin/<target>` and creates the new branch from it, falling back to the
    local `<target>` ref when origin is unavailable.
  - Resumes an existing local or remote branch by checking it out instead of recreating it.
  - Uses `--no-track` so the branch stays upstream-less until `hardis:work:save` pushes it
    with `-u`, matching the previous behavior.
- Add worktree helpers `listGitWorktrees()` (parses `git worktree list --porcelain`) and
  `getBranchWorktreePath()`.
- `getGitDeltaScope` (in `common/utils/gitUtils.ts`): when the local target ref cannot be
  updated (e.g. it is checked out in another worktree), fall back to fetching and computing
  the delta against `origin/<target>`.

## i18n

Adds three keys across all 9 locale files: `branchAlreadyCheckedOutInWorktree`,
`unableToFetchTargetBranch`, `unableToFindTargetBranchToBranchFrom`.

## Notes

The new branch no longer pulls into a local target ref, so the target branch is never
mutated in the current worktree. Behavior is unchanged for users not using worktrees:
the branch is still created from the latest origin state of the target.

This solves #1946


nvuillam (approved):
Thansk a lot for your PR :)

i made a few test and completed your PR so nothing breaks in case of worktree presence, don't hesitate to have a look on my commits :)

nvuillam:
Fixes #1968

PR #1969, opened 2026-06-24
fix: correct inverted notification types in org monitor errors

yamioliva:
This PR fixes an inverted notification type assignment in `src/commands/hardis/org/monitor/errors.ts`.

Current behavior:
- Apex notification is sent with `type: 'FLOW_ERROR'`
- Flow notification is sent with `type: 'APEX_ERROR'`

Expected behavior:
- Apex notification should use `type: 'APEX_ERROR'`
- Flow notification should use `type: 'FLOW_ERROR'`

## Changes made

- updated `buildApexNotification()` to return `type: 'APEX_ERROR'`
- updated `buildFlowNotification()` to return `type: 'FLOW_ERROR'`

## Impact

This change does not modify how Apex and Flow errors are collected.
It only corrects the notification labeling for downstream consumers such as monitoring integrations and Grafana-based filtering/routing.

### DIFF
--- a/src/commands/hardis/org/monitor/errors.ts
+++ b/src/commands/hardis/org/monitor/errors.ts
@@ -572,7 +572,7 @@ In agent mode:
         : t('monitorErrorsApexNone', { org: orgMarkdown, days: this.days });
 
     return {
-      type: 'FLOW_ERROR',
+      type: 'APEX_ERROR',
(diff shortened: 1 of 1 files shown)


nvuillam (approved):
Thanks a lot for your PR @yamioliva 😃

PR #1973, opened 2026-06-26
feat: add hardis:doc:data-dictionary command

TiercelinQ:
## What

New command `hardis:doc:data-dictionary`: exports an Excel (.xlsx) data dictionary for one or more Salesforce objects, read live from the target org.

## Workbook structure

- `Index` sheet: one row per object (API name, label, custom flag, key prefix, field / validation-rule / record-type counts)
- One fields sheet per object: API name, label, type, required, unique, external id, length/precision, reference, picklist values, default value, formula, help text, description, custom
- Consolidated `Validation Rules` sheet, with the error condition formula (read from the Metadata API)
- Consolidated `Record Types` sheet (single SOQL query)

## Notes

- Object selection via `--objects` (comma-separated) or an interactive prompt. `--objects` is required in `--agent` / CI mode.
- Picklist values are capped at 50 with an overflow note.
- New `forceTextColumns` option on the shared CSV-to-XLSX helper so identifier columns with leading zeros (e.g. key prefixes "001") keep their text form instead of being auto-typed to numbers.

## Tests

- Unit tests for picklist formatting and the forceTextColumns round-trip.
- yarn compile and yarn lint pass; manual run against a sandbox (Account + Contact) produces the 5 expected sheets.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


nvuillam:
@TiercelinQ it looks great, thanks for the PR :)

A few updates / questions:

1. Excel seem to think the file is broken, is there are way to avoid it ?
2. 
image

3. Can you fix the cell height ? (ex: picklist value for Type in the screenshot below)
image

4. Can you sort the fields by alphabetical order ?

5. (optional) Can we have hyperlinks from Index tab to other tabs + validation rule & record type tabs ? And maybe a link "back to index" a little below the end of these pages, to ease the navigation in case there are 50 tabs ?

Many thanks :)

nvuillam (changes requested):
Amazing, just a few changes and we're good to merge, thanks a lot :)

nvuillam:
@TiercelinQ i updated your branch as i fixed a CVE issue in another PR :)

nvuillam (approved):

nvuillam:
Finally i made the fix, will be released ASAP , thank for the PR :)

PR #1974, opened 2026-06-26
feat: add hardis:mdapi:read and hardis:mdapi:upsert (CRUD-based Metadata API)

SalikLP:
Two standalone, manually invoked commands that read and write metadata through
  the CRUD-based Metadata API, as an alternative to file-based deploy/retrieve for
  types that come back incomplete without source tracking e.g. (Profile, PermissionSet).

-   hardis:mdapi:read: reads chosen components whole (readMetadata) and writes complete source files.
-   hardis:mdapi:upsert: pushes local source components whole (upsertMetadata), with an interactive whole-component-overwrite confirmation.
- Shared util uses only public @salesforce/source-deploy-retrieve APIs plus fast-xml-parser and fast-xml-builder.
- Use the @salesforce/source-deploy-retrieve RegistryAdapter to determine if Metadata read/upsert is supported by the CRUD API. Will give a warning if unsupported but otherwise continue.
- Chunked at 10 per call (200 for CustomMetadata/CustomApplication), with a validated --chunk-size override; aggregates per-component results and exits non-zero on failure unless --ignore-errors.
- Deliberately not wired into the deploy:smart pipeline (no tests/validation/coverage). i18n added to all 9 locales; unit tests for the pure logic.


nvuillam (commented):
I think we don't need the files in docs/ :)

SalikLP:
Haha you are absolutely right. 
I had a lot of balls in the air when I submitted this 😄 I have removed them and force pushed a new commit

nvuillam:
@SalikLP please can you fix the megalinter issues ? :)

nvuillam (approved):
Great :)

nvuillam:
@SalikLP it' released, u can now play with the UI part 😊😊

PR #2006, opened 2026-07-16
fix: Flow git diff not working when project does not use force-app package directory

0ptaq0:
`MetadataUtils.findMetaFileFromTypeAndName` looks for metadata files only in a hardcoded `force-app` folder. On projects using custom package directories (e.g. `src-dx`), the Flow file lookup always returns null and the PR flow diff crashes with:

```
[FlowGitDiff] Unable to generate Flow diff for X: The "path" argument must be of type string. Received null
```

Changes:
- read `packageDirectories` from `sfdx-project.json` (fallback to `force-app` when missing/empty)
- skip the flow diff with a warning when the flow file cannot be found, instead of throwing on a null path
- new i18n key in all locales + unit tests

Tested on a real GitLab CI project with a `src-dx` package directory - the MR now gets the mermaid flow diff comment. Projects using `force-app` behave exactly as before.


nvuillam (approved):
Well played, thanks for the PR :)

i'll handle trivy issues in another PR before releasing ^^

0ptaq0:
No problem Nicolas, open source is not only about to take but also about giving. It ain't much but it is honest work 🤣

PR #2020, opened 2026-07-24
feat: capture real deployment metrics from smart deploy

mcarvin8:
## Summary

`smartDeploy()` now parses the JSON output of the `sf project deploy start` / `sf project deploy quick` calls it already makes, and accumulates real numbers (components deployed/total/failed, tests run/failed/total, code coverage percent, duration, quick deploy / delta / success flags) instead of the hardcoded zero stub that was previously sent in the `DEPLOYMENT` notification's `metrics` payload.

Those metrics flow through the existing `NotifProvider` -> API provider pipeline, which already pushes to Prometheus or InfluxDB when `NOTIF_API_METRICS_URL` is configured. No new persistence or endpoint was added, this only fills in numbers that were always zero.

### How this differs from `org diagnose deployments`

The existing `org diagnose deployments` command queries the org's Tooling API for `DeployRequest` records after the fact. That works for anything the org tracks server side, but it can't see things the CLI does at deploy time, like package installs, quick deploy usage, or per-invocation duration. It's also a pull, run on a schedule against org history, so it does not produce a data point per pipeline run.

This change is a push, emitted once per `smart deploy` invocation, from data the CLI already has in hand. It's meant to complement `org diagnose deployments`, not replace it: the org-side command stays the source of truth for anything that happened in the org regardless of how it got deployed, while this gives per-pipeline-run granularity (including quick deploys and delta info) that the Tooling API can't provide.

Only `project deploy smart` was wired up, since it's the command CI/CD pipelines actually run. `project deploy start`, `project deploy quick`, and `project deploy notify` still call the same notification helper without the new metrics and fall back to zero, unchanged from current behavior.

## Test plan

- [x] `yarn compile`
- [x] `yarn lint`
- [ ] Run `project deploy smart` against a real org and confirm `NOTIF_API_METRICS_URL` receives non-zero component/test/coverage values


nvuillam (approved):
Thanks a lot again ;)

PR #2033, opened 2026-08-01
Handle Flow deletion in destructive changes via Tooling API

SalikLP:
# Delete Flows listed in destructive changes

`hardis:project:deploy:smart` can now delete a Flow that appears in a destructive manifest. Until now those members either failed the deployment or were silently ignored, and the Flow had to be removed by hand in the org.

Flow members are stripped from the manifest sent to the org and deleted through the Tooling API after the deployment succeeded. Nothing about Flow deletion travels through `sf project deploy`.

## What it does

For every `Flow` member found in `destructiveChanges.xml` / `preDestructiveChanges.xml`:

1. Existence check in the target org.
2. Deactivate the Flow (`FlowDefinition.activeVersionNumber = 0`, Tooling API).
3. Query the Flow Interviews blocking the deletion.
4. Delete those interviews, **only** if authorized (see below).
5. Delete relevant Flow version.
6. Verify.

Details:

- A bare member (`MyFlow`) deletes the whole Flow, versions enumerated live from the target org. A versioned member (`MyFlow-3`) deletes that version only, and only deactivates the Flow if version 3 happens to be the active one, so trimming an obsolete version never takes a live Flow offline.
- A wildcard (`*`) is refused.
- Every step is idempotent, so the whole sequence is re-runnable: the existence check is a pure read, deactivating an inactive Flow is a no-op, deleting an absent version counts as success.
- Deactivation is unconditional and comes before the interview gate. That ordering is what makes a blocked Flow converge: once deactivated, no new interview can start, so the blocking population can only shrink and a plain pipeline retry eventually succeeds. Gating first would leave a Flow active whenever the opt-in is off and let interviews accrue forever.
- Flows are processed independently and the worst status decides the exit, so one blocked Flow does not stop the others. Per-Flow markers: `FLOW_DELETE_BLOCKED` (deactivated, versions still blocked by interviews) and `FLOW_DELETE_NOOP` (nothing deletable in this org).
- A `--check` run does the read-only part: it reports the deletion plan in the Pull Request comment and fails immediately if interviews block a deletion, instead of burning a full validation first.

## Deleting Flow Interviews is opt-in

Deleting interviews destroys in-flight process state and is irreversible, so it needs explicit consent through `FLOW_DELETE_INTERVIEWS`:

- Pull Request description keyword (standalone line)
- `FLOW_DELETE_INTERVIEWS` env variable, with the optional `_<branch>` variant
- `flowDeleteInterviews: true` in `.sfdx-hardis.yml`

Wired through `customBehaviors` like `purgeFlowVersions`. Off by default: without it, a blocked Flow reports `FLOW_DELETE_BLOCKED` and names the interviews to resolve.

## Reporting

Console table, CSV report, and a block in the Pull Request comment listing each Flow, its scope, its previous active version, the versions deleted, the interviews deleted and the final status. The pre-deactivation active version number is logged as compensation information and never re-applied automatically.

## Design history: multiple approaches tried and dropped

**Attempt 1: one deploy does everything.** Deploy a `FlowDefinition` with `activeVersionNumber = 0` in the constructive package and delete the versions as post-destructive changes in the same `sf project deploy`, betting that Salesforce processes the package before the deletes. It validates under `--check` and then fails on the real deploy: Salesforce tries to re-point the active version after the versions are gone, giving `UNKNOWN_EXCEPTION: System.security.NoDataFoundException`. A Flow cannot be deactivated and have its active version deleted in one deployment.

**Attempt 2: deactivate out-of-band, keep the deletes in the deploy.** Deactivate through the Tooling API just before the deploy, let the deploy delete already-inactive versions. Real deploys worked. It broke on the two remaining constraints:

- A `checkOnly` deploy never commits the deactivation, so the destructive members can never validate. The workaround was injecting a generated deactivating `FlowDefinition` into the validated package, which made the validated container differ from what the real deploy sends.
- A quick deploy replays the exact validated container. A deletion left out of the validation is silently dropped from the real deployment, and one left in cannot validate. Both directions lose.

**Attempt 3, reactive interview handling.** Run the deploy, and only on an interview-block failure delete the blocking interviews and retry. Dropped with the in-deploy design. Two things it established are still in use: the `FlowInterview.FlowVersionViewId` to `FlowVersionView.DurableId` to `FlowDefinitionView.ApiName` join (with the two Salesforce quirks, `FlowVersionView.Id` being the placeholder `000000000000000AAA` on every record, and `FlowInterview.FlowVersionViewId` holding the 15-char form of the 18-char `DurableId`), and the fact that a deploy error names only about 10 blocking interviews, which is why the gate queries the org instead of parsing the error.

Moving the deletion out of the deploy resolves all of it at once and removes a body of workaround machinery: destructive-manifest version expansion, generated `FlowDefinition` source files with their backup/restore, `package-no-overwrite` survival re-injection, the reactive retry loop, and the `--check` interview-block false-positive classifier. Version enumeration happens live against the target org, which is more correct: version counts differ across int / uat / prod, so a list computed at delta-generation time might be wrong.

## Fixes included

- `packageXmlToDeploy`, `packageXmlToDelete` and `packageXmlToDeletePreDeploy` (and their `PACKAGE_XML_TO_DEPLOY`, `PACKAGE_XML_TO_DELETE`, `PACKAGE_XML_TO_DELETE_PRE_DEPLOY` env variables) were ignored. `?:` binds looser than `||`, so the configured path was only tested for truthiness and then discarded, and the default `manifest/` or `config/` path always won. Worst case, a project pointing at a custom destructive manifest deleted nothing and still exited 0. Same bug in `deploy/smart.ts`, `deploy/sources/metadata.ts` and `orgUtils.ts`, all pre-existing on main.
- A configured destructive manifest that does not exist on disk now logs a warning instead of being skipped silently.

### Fixes around purge:flow
- - `bulkDeleteTooling` rewritten. `conn.tooling.destroy(type, ids[])` routes to the sObject Collections endpoint, which does not exist in the Tooling API and always 404s, so every call fell back to one `sf data:delete:record` process per record. It now batches DELETE subrequests through the Tooling composite endpoint, 25 per call, with the one-by-one path kept as a fallback. This should make the `hardis:org:purge:flow` faster in case of many FlowVersions to delete.
- `hardis:org:purge:flow` now surfaces the real error message in the `Error` column of the report.
- Interview scoping (manageDeleteFlowInterviews) - only interviews on versions whose delete failed with the Flow Interview -  marker are targeted. Before: ids regex-scraped from every error. Now: query FlowInterview by the blocked version ids (queryFlowInterviewIdsForFlows), error-parsing kept as fallback, ids deduped across 15/18-char forms.
- Interview delete results normalized - Bulk v2 returns sf__Id and no success flag; mapped to {Id, success, errors}. Also push(array) -> push(...), so results are no longer nested and no longer render as errors.
- Summary count fixed - this.deletedRecords is array-of-arrays; count now uses .flat().
- Result table - new Type column (Flow Version / Flow Interview); details read from new flowVersionDetailsById / flowInterviewDetailsById caches keyed on the 15-char id, instead of this.flowRecords (which gets reset mid-run -> blank rows); Flow Status -> Status; error cell falls back to errors[].message.
- Interview details captured before deletion - displayFlowInterviewToDelete now selects Id and caches rows; post-delete they can't be queried.

## Tests

- Unit tests for the new `flowDeletionUtils` module, covering manifest parsing and stripping, wildcard refusal, version expansion, the interview query and its 15/18-char Id handling, the opt-in resolution channels, chunking, and the outcome/report shaping.
- Manual campaign of 33 scenarios against a sandbox. Covered: whole Flow and versioned members, active vs inactive version, wildcard refusal, absent Flow, replay/idempotency, opt-in on and off across all channels, quick deploy, mixed and Flow-only manifests, pre- and post-destructive independence, delta deployment, deploy failure short-circuit, and the report content in two locales. The two defects the campaign found in this branch (the CSV report losing the pre-destructive phase, and the `packageXmlToDelete` precedence bug) are also fixed here.


nvuillam:
## Review: relevance, quality, backward compatibility

Reviewed the full diff, then installed dependencies on the branch and ran typecheck, lint and the test suite locally.

Overall: relevant and well built. Two CI blockers, both small and diagnosed below, plus three behavior changes that should be called out explicitly in the CHANGELOG before merge.

### Relevance

Deleting a Flow through a metadata deployment is close to impossible in Salesforce (bare member fails, active version cannot be deleted, `checkOnly` never commits a deactivation, quick deploy replays the validated container). Taking it out of the deploy and into Tooling API calls is the right call, and the design history in the PR description matches what the code ended up doing. The bundled fixes (`packageXmlToDelete` precedence, `bulkDeleteTooling`, `purge:flow` interview scoping and result table) are real bugs, not scope creep.

### Quality

Verified locally on `feature/destructive-flow`:

- `tsc --noEmit` clean, `yarn lint` clean.
- Full suite: 787 passing, 0 failing in a no-color terminal (see blocker 1).
- `flowDeletionUtils.ts` at 95.45% statement coverage per the CI report.
- All 53 new i18n keys present in all 9 locale files, none left as untranslated English (the few identical to `en` are "Flow" and "Status", correctly identical).
- The `deferSuccessPullRequestComment` plumbing patches all four `managePostPullRequestComment` sites in `smartDeploy`, including the three early-return branches, so a Flow-only destructive PR (empty package.xml after stripping) still routes its comment correctly.
- Ordering in `run()` is right: `prepareFlowDeletion` runs before `setAdditionalOptions`, so the `DESTRUCTIVE_CHANGES_AFTER_DEPLOYMENT` extra command picks up the stripped manifest rather than the original.

### CI blockers

**1. Two new tests fail in all 6 unit-test jobs. Root cause is chalk ANSI codes, not the code under test.**

`uxLogTable` emits the table through `c.italic()` (`src/common/utils/index.ts:1534`). Chalk re-opens the style at the start of every line, so each row begins with `[3m`. `tableFlows()` does `line.trim().startsWith('HardisTest')`, which is false once colors are on. It passes locally because chalk detects no color support, and fails on GitHub Actions because it does.

Reproduced exactly:

```
FORCE_COLOR=1 npx mocha "test/common/utils/flowDeletionUtils.test.ts"
  2 failing
  AssertionError: expected [] to deeply equal [ 'HardisTestFlowOne', ...(1) ]
```

Fix: strip ANSI escapes in `tableFlows()` (`test/common/utils/flowDeletionUtils.test.ts:1183-1188`) before the `startsWith` filter. Note that `CI=true` alone does not reproduce it, `FORCE_COLOR=1` does.

**2. MegaLinter is red on cspell only**, every other linter passes. 17 unknown words: `subrequest` / `subrequests` in `apiUtils.ts:255,260` and in the test, plus the fake Id fragments `Gdnw`, `Hotx`, `Iabc`, `gdnw` in the test. Either add them to `.github/linters/.cspell.json` or use spell-clean mock ids.

### Backward compatibility

Three behavior changes for existing pipelines. None are wrong, all deserve a release note.

**Loss of deploy atomicity for Flow destructive members.** A Flow in `preDestructiveChanges.xml` used to be deleted inside the `sf project deploy` transaction, so a failed deploy rolled it back. It is now deleted and committed before the deploy runs, so a deploy that then fails leaves the Flow gone and the new metadata not deployed. The deactivation is likewise committed and never reverted. This is documented in the command help and the retry story is sound (deactivating is what makes a blocked Flow converge), but the failure semantics changed.

**`--check` no longer validates Flow destructive members against the org.** A typo'd or nonexistent Flow member used to surface at validation time; it now reports `FLOW_DELETE_NOOP` and passes. The rationale (the same delta is replayed along the promotion chain and may already have been applied) is reasonable, but it does remove a typo safety net.

**The `packageXmlToDelete` precedence fix can change which metadata gets deleted.** This is the most consequential item. Before, `A || B || C ? x : y` meant a configured `packageXmlToDelete: config/x.xml` was only tested for truthiness and then discarded, and `manifest/destructiveChanges.xml` won. After the fix the configured path wins. For a project that sets the config *and* has `manifest/destructiveChanges.xml` on disk, the first run after upgrading deletes a different set of metadata. Same shape for `packageXmlToDeploy`, and for `--packagexml`, which smart deploy was silently ignoring. I checked the rewritten ternaries in `metadata.ts` and `orgUtils.ts`: the fallback chains stay equivalent when nothing is configured, so only the configured case moves.

Lower impact: `bulkDeleteTooling` is exported from the public `sfdx-hardis/plugin-api` and its return shape is now always `{results: [{Id, success, errors}]}`. In practice the old happy path always 404'd and fell through to the same `{results}` shape, so real breakage is near zero, but it is a public API change.

Also worth noting: the `generateReportPath` fix in `bulkUpdate` moves every bulk CSV report from a bogus relative filename into the reports directory. That is a clear improvement, just a path change.

### Correctness notes

1. **`isAlreadyDeletedError` can mask a systematic failure.** It treats `NOT_FOUND` and `does not exist` as success. If the composite endpoint or API version were ever wrong, every subrequest returns 404 and every version is counted as deleted. Whole-Flow deletions are protected by the leftover-verification query at the end of `deleteOneFlow`, but **versioned members have no verification step**, so a false `SUCCESS` is reachable there. Either verify versioned deletions too, or narrow the match so only a delete that actually reported `ENTITY_IS_DELETED` counts as already gone.

2. **`FLOW_DELETE_MAX_ATTEMPTS` and `FLOW_DELETE_RETRY_DELAY_MS` are undocumented** (not in the command description, not in the JSON schema). `Number('abc')` also yields NaN, which makes `for (attempt = 1; attempt <= NaN; ...)` run zero times and report a confusing `FLOW_DELETE_ERROR: versions remaining` without a single delete attempt. A value of `0` does the same. A guard plus a doc line would help.

3. **SOQL IN-clause growth.** `queryFlowInterviewIdsByFlow` puts both the 15 and 18 char form of every version DurableId into the `FlowInterview` IN clause. Many Flows times many versions can approach the 100k character SOQL limit. Chunking that clause would be safer.

4. Flow names are interpolated into SOQL without escaping. Flow API names cannot contain quotes so it is not exploitable, but the escape is cheap.

5. The comment above `deleteFlowVersionsViaToolingApi` says versions are deleted "oldest version first, so each version keeps its own error attribution", while the implementation sends one composite batch. Harmless, just stale.

### Missing per repo convention

- **No `CHANGELOG.md` entry** under `## [beta] (main)`. Given the three behavior changes above, this one matters.
- **`docs/schema/sfdx-hardis-json-schema-parameters.html` not regenerated** for the new `flowDeleteInterviews` property. 6 of the last 8 commits touching `config/sfdx-hardis.jsonschema.json` regenerated it in the same commit.
- `docs/hardis/project/deploy/smart.md` is stale relative to the new command description, but this repo regenerates command docs in dedicated "build doc" commits, so that one is fine to leave.

### Minor

The branch is based on v7.23.0 and is 3 commits behind `main`. It does not touch `package.json` or `yarn.lock`, so rebasing will not revert the recent dependency bumps.

---
_Generated by [Claude Code](https://claude.ai/code)_

SalikLP:
@nvuillam let me know if there is anything you would like to have me fix and/or change 🙂

nvuillam:
@SalikLP you can start by fixing ci issues ;) (probably merge main in your branch first ^^)

then update changelog 

then these ^^

```
FLOW_DELETE_MAX_ATTEMPTS and FLOW_DELETE_RETRY_DELAY_MS are undocumented (not in the command description, not in the JSON schema). Number('abc') also yields NaN, which makes for (attempt = 1; attempt <= NaN; ...) run zero times and report a confusing FLOW_DELETE_ERROR: versions remaining without a single delete attempt. A value of 0 does the same. A guard plus a doc line would help.

SOQL IN-clause growth. queryFlowInterviewIdsByFlow puts both the 15 and 18 char form of every version DurableId into the FlowInterview IN clause. Many Flows times many versions can approach the 100k character SOQL limit. Chunking that clause would be safer.

Flow names are interpolated into SOQL without escaping. Flow API names cannot contain quotes so it is not exploitable, but the escape is cheap.
```

nvuillam:
@SalikLP I'm not far from validating the PR :)
But there is too many flow-related code in smart.ts, please can you move a big part of it elsewhere ? As smart.ts is supposed to be an orchestrator, it must remain not too hard to read, and mostly redirect to other classes/methods :)

nvuillam (approved):
Great, thanks :)

PR #2038, opened 2026-08-06
feat(diagnose:audittrail): add monitoringAllowedUsersActions to allow expected actions per user

pranayjswl007:
## Problem

`hardis:org:diagnose:audittrail` currently offers two granularities for tuning suspect detection, and there's a gap between them:

- `monitoringExcludeUsernames` hides **everything** a user does (SOQL-level exclusion)
- `monitoringAllowedSectionsActions` allows an action for **every** user

A common real-world case fits neither: an integration user that provisions Experience Cloud / portal users, where Salesforce auto-creates account roles ("<Account> Customer User") logged as `Manage Users / createdrole` under the integration user - dozens per month of permanent noise. Excluding the whole user hides any genuinely suspicious action it might ever perform; allowing `createdrole` globally hides *human* role creation, which is a real signal since roles affect sharing.

## Solution

New `.sfdx-hardis.yml` property `monitoringAllowedUsersActions`: a map of username to the actions that are expected from that user. Listed actions are reported as not suspect for that user only; any other action from the same user - and the same action from any other user - is still flagged.

```yaml
monitoringAllowedUsersActions:
  "provisioning-user@cloudity.com": ["createdrole"]  # only createdrole is expected from this user
  "another-integration@cloudity.com": []             # empty list = all actions allowed (records stay visible in the report, unlike monitoringExcludeUsernames)
```

The `[]` semantics mirror `monitoringAllowedSectionsActions`. The check also covers the Custom Settings audit (action name `customSetting<ApiName>`).

## Changes

- `src/commands/hardis/org/diagnose/audittrail.ts`: read the new config property, early-allow matching user+action pairs in `queryAuditTrail`, and skip matching rows in `handleCustomSettingsAudit`; command description documents the property (generated docs will pick it up on release)
- `config/sfdx-hardis.jsonschema.json`: schema entry for the new property
- `CHANGELOG.md`: beta entry

## Validation

Ran the modified build against a production org exhibiting exactly this pattern (2-day window):

- **Without the config** (control, same build): the integration user's 4 `createdrole` records flagged as suspect
- **With the config**: the same 4 records present in the report with `Suspect=false`, while 43 suspect records from other users remain flagged, including `createdrole`-adjacent Manage Users actions

`tsc --noEmit` and `eslint` pass with no new findings.


pranayjswl007:
Hi maintainers, could you help merge this one?

MegaLinter was failing on `cspell` due to `createdrole` (introduced by this PR) not being in the custom dictionary — fixed that in the latest commit.

The check is now red for a different reason: `osv-scanner` and `trivy` are flagging a HIGH severity CVE in `js-yaml` (`GHSA-5p4m-2wfm-xmqj`) via `yarn.lock`. This isn't caused by this PR — `yarn.lock`/`package.json` are unchanged vs. `main`, and the same failure would occur on `main` and any other open PR right now (looks like a newly published advisory that a fresh vulnerability DB pull just started catching). Happy to bump `js-yaml` here if that's preferred, but figured it's probably better handled as a separate repo-wide dependency update. Let me know how you'd like to proceed.

nvuillam (approved):
Good idea :)
i'll merge once the CI is good ^^ (i updated your branch to solve the issues not related to your PR)

PR #2073, opened 2026-08-19
Fixed agentmode not being respected on files import/export

SalikLP:
I got hit by a prompt even when running with --agent when trying to import some files in a script


nvuillam (approved):
Good catch as always, thanks :)

PR #2117, opened 2026-08-26
Prevents Azure report publish task failure

nvuillam:
Addresses an issue where Azure Pipelines (both check and deployment jobs) would fail on the 'Publish sfdx-hardis reports' step.

This failure occurred because the `hardis-report` folder is not created when there are no changes to deploy (e.g., an empty delta package.xml), causing the publish task to hard-fail on a missing path.

Adds `continueOnError: true` to the `publish` task for `hardis-report` in both the `azure-pipelines-checks.yml` and `azure-pipelines-deployment.yml` templates. This ensures that pipelines can complete successfully even if the report folder is absent.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 ## [beta] (main)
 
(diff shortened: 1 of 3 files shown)


Sakthes:
Great thank you for this fix.

Issue #2152, opened 2026-09-02
[sf hardis:org:test:apex] add option to pass custom wait timeout


arafesthain:
Hello Hello, 
default value for timeout (60min) is too low for certain org can we add the --wait option on the cli ?

```
$ sf hardis:org:test:apex --wait 120
Error (2): Nonexistent flag: --wait
```

+ It will be nice to surface all underlying sf cli parameters to the hardis cli?

nvuillam:
@arafesthain you can already override the value using variable `SFDX_TEST_WAIT_MINUTES `

See command doc -> https://sfdx-hardis.cloudity.com/hardis/org/test/apex/

If you want to add an explicit --wait to the hardis command, please submit a pull request and i'll accept it :)

arafesthain:
Hey here we go :) #2159

PR #2159, opened 2026-09-03
feat([GH-2152]) : boilerplate unknown hardis extension arguments to the sf cli

arafesthain:
#2152


nvuillam:
@arafesthain this is too generic, please can you explicitely declare in apex.ts arguments that can be forwarded to sf apex test run ?

arafesthain:
@nvuillam I tought the same at the beginning but sf cli is here to reject whatever is not ok. (SOC)

nvuillam:
You can find the définitions here :)

https://github.com/salesforcecli/plugin-apex/blob/main/src%2Fcommands%2Fapex%2Frun%2Ftest.ts

arafesthain:
Yep yep I know. Will replace hardis command by sf one will be fine. I Dont see the point of replicating code that sf cli is doing

PR #2169, opened 2026-09-04
feat(test-cases): write test notebooks and upsert them to Azure DevOps, ServiceNow and Xray

SamuelBouleau:
## What this adds

Two commands under a new `hardis:ticket:test-cases` topic, and the parsing, rendering and provider code they share. Neither command needs an org or a Salesforce project.

| Command | What it does |
|---|---|
| [`hardis:ticket:test-cases:init`](https://sfdx-hardis.cloudity.com/hardis/ticket/test-cases/init/) | Writes the test cases of a ticket into a notebook a tester fills in: a formatted Excel workbook, a CSV, or a markdown table |
| [`hardis:ticket:test-cases:upsert`](https://sfdx-hardis.cloudity.com/hardis/ticket/test-cases/upsert/) | Reads a notebook back and creates or updates its test cases in Azure DevOps, ServiceNow Test Management 2.0 or Xray Cloud |

The intended flow: an agent (or a script) drafts the cases as JSON, `init` turns them into a workbook, a human corrects it in Excel, and `upsert` sends the corrected version to the tracker, from a laptop or from a CI job. The last two steps need no model and no LLM token.

## `hardis:ticket:test-cases:init`

- **Input**: `--testsjsonfile` (a `NormalizedTestCase[]` JSON file, validated field by field) or `--notebook` (an existing `.md`, `.xlsx` or `.csv`). Without either, it writes a blank notebook with numbered identifiers, from `--kind`, `--ticket-number`, `--modules` and `--rows`, prompting for the kind and the ticket when they are missing (the ticket is proposed from the current branch name).
- **Output**: `--format xlsx` (default), `csv`, `md` or `both` (xlsx + CSV), path forced with `--outputfile`. Writing over the notebook being read is refused, so a tester's results are never erased.
- **Notebook kinds**: the identifier suffix decides the column set: `PROJ-123-F01` functional (with SOQL query and steps), `PROJ-123-T01` technical (with class / method under test), `PROJ-123-01` TMA (no module, no priority). `--kind` forces it. An unreadable identifier raises instead of guessing, and a payload mixing kinds is refused unless `--kind` is set.
- **Workbook**: bold frozen header, `Résultat obtenu` / `Commentaire` / `Statut` left empty for the tester, `Statut` restricted to `OK, KO, N/A, Bloqué`, and a `Synthèse` sheet with counts per module and priority.
- **CSV**: `;` delimiter, UTF-8 BOM and CRLF so Excel opens it cleanly on a double click, with a `SYNTHÈSE` footer the reader stops at. Quoted cells holding line breaks are read back correctly.
- **Formula guard**: a cell starting with `=`, `+`, `-` or `@` gets an apostrophe prefix on write (footer and blank template included), removed again on read.
- `--agent` skips the prompts: `--kind` and `--ticket-number` are then required for a blank notebook.

## `hardis:ticket:test-cases:upsert`

- **Input**: `--notebook` or `--testsjsonfile`, exactly one.
- **Idempotency**: every case carries `TESTKIT:<TICKET>:<SHORT ID>` (a tag on Azure DevOps, a `short_description` prefix on ServiceNow, a label on Xray), so a second run updates instead of duplicating.
- **Refuses an unfinished notebook** as a whole: `À COMPLÉTER` markers, unresolved `{PLACEHOLDER}` tokens, stringified `null` / `undefined`, empty required fields, duplicated identifiers. Every offender is listed.
- **Providers activate from environment variables** already used elsewhere in sfdx-hardis; `--provider` only narrows:
  - Azure DevOps: `SYSTEM_COLLECTIONURI`, `SYSTEM_TEAMPROJECT`, `CI_SFDX_HARDIS_AZURE_TOKEN` or `SYSTEM_ACCESSTOKEN`
  - ServiceNow: `SERVICENOW_URL`, `SERVICENOW_USERNAME`, `SERVICENOW_PASSWORD`
  - Xray: `XRAY_CLIENT_ID`, `XRAY_CLIENT_SECRET`, `JIRA_HOST`, `JIRA_PROJECT_KEY`, `JIRA_EMAIL`, `JIRA_TOKEN`, optional `XRAY_REGION`
- **Best effort per case**: exit code 0 when everything went through, 2 when some cases failed, 1 when no provider could be used. A CSV / XLSX report lists provider, case, action, tracker id, URL and error for every case.
- `--dry-run` probes the providers and looks the cases up without writing. An interactive confirmation is asked before writing, skipped with `--agent` or in CI.
- **Per provider**:
  - Azure DevOps: a `Test Case` work item in the current project, with `Microsoft.VSTS.TCM.Steps`, inheriting area path, iteration and assignee from the story work item (the number is read from the ticket key), linked with `TestedBy-Reverse`. Not attached to a Test Plan or Suite.
  - ServiceNow: records in `sn_test_management_test`, `sn_test_management_test_version` and `sn_test_management_step`. Updates change the test record only, never the steps, and nothing is linked (no generic relation table).
  - Xray Cloud: `createTest` GraphQL mutation with its steps, then a Jira `Test` issue link. Updates go through Jira REST v2 (summary, description, priority, labels), not the steps.

## Docs, i18n, dependencies

- Generated command pages `docs/hardis/ticket/test-cases/init.md` and `upsert.md`, entries in `docs/commands.md`, `mkdocs.yml` and the agent commands table of `docs/salesforce-agentic-automation.md`.
- New i18n keys in the 9 locales. Notebook domain words added to `.github/linters/.cspell.json`.
- No new dependency and no new config property: `exceljs` and `azure-devops-node-api` are already direct dependencies.

## Tests

Unit tests for the markdown, xlsx, CSV and JSON readers, the renderer (including write then read round trips on the three formats), the guards, the template rows, the dispatch logic and each provider (Azure DevOps through an injected API stub, ServiceNow and Xray through the fetch stub).

## Changes after review

- Merged `main` (CHANGELOG and cspell conflicts).
- **Xray**:
  - issue lookup calls `/rest/api/3/search/jql`, since Atlassian removed `/rest/api/3/search`
  - updates go through REST v2, because v3 rejects a plain string description
  - a bare `JIRA_HOST` gets `https://`
  - every call has a 60 s timeout
- **ServiceNow**: steps go to `sn_test_management_step` (with the text in `step`); `sn_test_management_test_step` does not exist in Test Management 2.0. The other step field names (`test_version`, `order`, `expected_result`) were not checked against a live instance.
- **Azure DevOps**:
  - step text is escaped twice as the Steps field expects, so `<Account Name>` no longer vanishes
  - steps with an expected result are typed `ValidateStep`
  - the lookup is scoped to the current project
  - line breaks show in the description
- **Dispatch**:
  - a case created but not linked is reported once, as failed, with its tracker id
  - exit code 1 when no provider passes its probe
  - the report keeps every column, error included
  - the summary goes through i18n
  - an action line follows the confirmation prompt
- **Notebook round trip**:
  - the formula guard apostrophe is removed on read, and the guard is reversible
  - `-1 ...` and `**bold**` step actions are no longer altered
  - pipes are kept (escaped in markdown)
  - `<br>` in text fields is decoded on read
  - quoted multi-line CSV cells are read correctly
  - duplicated ids are refused
- **init**:
  - `--format both` on a blank notebook writes the CSV too
  - the command refuses to overwrite the notebook it reads
- Removed three i18n keys left over from the earlier `render` / `template` commands.

## Open points for the maintainer

Found in review, not changed here because each one changes behavior or scope:

1. **Provider activation.** The Azure DevOps provider turns on from `SYSTEM_COLLECTIONURI` + `SYSTEM_TEAMPROJECT` + `SYSTEM_ACCESSTOKEN`, which the Azure Pipelines templates already map, and the ServiceNow one from the ServiceNow ticketing variables. A Jira + Xray team on Azure Pipelines therefore also writes test cases to Azure DevOps unless `--provider` is passed. Should activation require an explicit opt-in?
2. **`--ticket-number` and the idempotency key.** A ticket override that differs from the id prefix yields a different key, so cases already pushed are created again, and the flag is ignored for `--testsjsonfile`.
3. **Missing columns read as defaults.** A TMA notebook has no priority or module column and a technical one has no steps, so an upsert from those notebooks resets priority to 2 and clears the module and steps on the tracker. The technical class / method is sent to no tracker.
4. **Steps are not updated on Xray and ServiceNow**, while the case is reported as updated. Xray Cloud GraphQL does offer step mutations.
5. **Azure DevOps token and project detection** do not reuse what `hardis:ticket:get` reads (`AZURE_DEVOPS_EXT_PAT`, git remote), and an update re-copies area, iteration and assignee from the story on every run.


nvuillam:
Hi @SamuelBouleau,

I merged `main` into your branch, then pushed a review commit (323efa506). Main changes:

- **English by default:** notebook headers, sheet names, statuses (Passed, Failed, Blocked, N/A), summary and the `TO BE COMPLETED` marker are now English, and the CSV uses `,`. The readers still accept the French headers and `;`, so your existing notebooks stay readable. The `tma` kind is now `maintenance` (`tma` is still accepted in JSON).
- **Fewer files:** `testNotebookTypes`, `testNotebookGuards` and `testNotebookInput` are merged into `testNotebookUtils.ts`, and `azureDevopsStepsXml` into `azureDevopsTestProvider.ts`.
- **Explicit tool choice:** `upsert` sends to one tool, chosen with `--provider` or the new `testCasesProvider` config property (with a prompt in interactive mode). It no longer guesses from environment variables, because the Azure pipeline templates set the Azure ones on every Azure CI job.
- **Providers:**
  - Azure DevOps: 404 answers from the SDK are handled, and the carrier work items are read before any write, dry run included. Updates keep the assignee, area, iteration and tags. `AZURE_DEVOPS_EXT_PAT` and the git remote are supported.
  - ServiceNow: the expected result goes in the step text, because TM 2.0 has no such column. A 200 answer that is not a record is refused.
  - Xray: JQL values are escaped, labels are added instead of replaced, and `jiraHost` can come from the config.
- **Safer upsert:** the idempotency key no longer changes with `--ticket-number`. Updates only send the columns the notebook has. `assertPushable` lists every problem at once.
- **init:** the default file name includes ticket, kind and date. An existing file is never overwritten silently. The format is taken from the `--outputfile` extension.

Could you test it locally before we merge?

```sh
git pull
yarn
yarn build
sf plugins link
```

Then run your usual scenarios with `sf hardis:ticket:test-cases:init` and `sf hardis:ticket:test-cases:upsert --provider <azure-devops|servicenow|xray>` (try `--dry-run` first) against your real Azure DevOps / ServiceNow / Xray instances.

- If everything works, please post a comment here saying your tests are successful.
- If something does not work, feel free to push new commits on this branch with the fixes.

Thanks!

SamuelBouleau:
## Local test report

Tested `323efa50` on Windows 11, Node 24.11.0, `@salesforce/cli` 2.150.6, plugin linked from source, against a real Azure DevOps project (`dsi-altis` / `DSI`) on a sandbox story. Two defects found and fixed in `4d5c44ba`, both verified again on the same project.

### Results

| Phase | Scope | Result |
|---|---|---|
| Build and unit tests | `yarn test:only`, `yarn lint`, `yarn build:doc` | 2232 passing, 0 failing, 0 ESLint error, no doc diff |
| `init` offline | 13 scenarios | 13/13 |
| Guards and `--dry-run` | 8 scenarios | 8/8, the interactive provider prompt was not reachable without a TTY |
| `upsert` writing to Azure DevOps | create, replay, manual edits, missing columns, rendering, locale, partial failure | 7/7 after the fixes |

### Confirmed on a live tracker

- Round trip xlsx to md to csv to xlsx is byte identical, including cells starting with `=`, `+`, `@` or `-`, `**bold**`, pipes, `=>`, quotes and multi line cells.
- Legacy header aliases and the `;` delimiter work on a real pre-English notebook in the three formats, tester columns preserved.
- The notebook guard lists every problem at once and stops before any tracker call.
- Explicit provider choice works through `--provider`, through `testCasesProvider` in config, and fails with exit 1 and the list of configured tools when neither is set in `--agent`.
- **Open point 2 closed**: the idempotency key no longer depends on `--ticket-number`. Cases pushed with a different `--ticket-number` matched their existing work items instead of being created a second time.
- **Old open point 5 closed**: an update no longer re-copies from the story. An unassigned test case stayed unassigned, a hand changed area path stayed, a tag added by hand was kept.
- The carrier is read before any write, dry run included: an unknown story number is refused with exit 1 and nothing is written.
- `<Account Name>` survives the double escaping, a step with an expected result is a `ValidateStep`, the story link is `TestedBy-Reverse`.
- Exit codes: 0 when everything passes, 2 on a partial failure with the tracker error in the report (`TF401324: Value for field 'System.Title' is too long.`), 1 when no provider can be used.

### Defects found, fixed in `4d5c44ba`

**1. The legacy TODO marker stopped blocking a push.** `TEST_CASE_TODO` is now `'TO BE COMPLETED'` only, so a notebook written before the English rendering, holding `À COMPLÉTER`, passed the guard and would have sent the placeholder to the tracker as the expected result. Those notebooks are exactly the ones the header aliases are meant to keep readable. Markers are now compared with case and accents folded, so both block the push. Verified: that legacy notebook is now refused with exit 1, naming the cell and the marker.

**2. An Azure DevOps update erased description blocks.** The priority, the steps and the tags are only sent when the notebook has the column, but `System.Description` was rebuilt from the notebook alone. Upserting a maintenance or technical notebook, which has no module and no SOQL column, removed the `Module:` and `SOQL query:` blocks from the work item, with no warning and no mention in the report. This is the other half of open point 3: the fields were protected, the description was not. The description now keeps the block of a field the notebook does not define, and drops it only when the notebook defines it as empty. Verified on the same work item: description byte identical before and after, priority, tags and steps untouched.

### Left to your judgment

**Legacy statuses are not mapped.** `OK`, `KO` and `Bloqué` are read back as is, then written into a cell whose validation list is `Passed, Failed, Blocked, N/A`, so a tester inherits cells that violate the dropdown. Mapping them on read would close it, but that is a behavior choice rather than a defect.

**Markdown bold is not converted in the step text.** `**Delete**` reaches Azure DevOps with its asterisks. Nothing is altered, so this may be intended, but a tester writing bold in a notebook sees the markup in the tracker.

**The Azure DevOps project comes from the git remote or the environment variables, not from the ticket key.** Run from a repository whose remote points at a different project, the upsert does not find the existing cases and creates duplicates there without saying anything. The project scoping you added behaves correctly, this is a documentation point, or a warning when the ticket key does not resolve in the target project.

### Not tested

ServiceNow and Xray: no instance and no credentials available here, so they stay covered by unit tests only. Open point 4, steps not updated on those two while the case is reported as updated, is therefore still open and unverified.

PR #2170, opened 2026-09-04
Implement REST deploy option for faster and supporting larger deploys.

SalikLP:
## Summary

Adds an option for `hardis:project:deploy:smart` to use Salesforce's Metadata REST API instead of the default SOAP Metadata API.

REST deployment can be enabled through either:

```yaml
# config/.sfdx-hardis.yml
useRestDeploy: true
```
or 

```bash
SFDX_HARDIS_USE_REST_DEPLOY=true
```

## Implementation
When enabled, the command sets Salesforce CLI's native SF_ORG_METADATA_REST_DEPLOY=true environment variable. Child sf project deploy processes inherit this value and select the REST transport.
This does not modify the user's Salesforce CLI configuration or persist the setting outside the command process.
The change also adds:
- JSON schema support for useRestDeploy
- Documentation and environment-variable reference updates
- Localized activation messages
- Unit tests for configuration precedence, Azure unresolved variables, disabled behavior, logging, and child-process inheritance
Validation
- TypeScript compilation passed
- ESLint passed
- JSON configuration and translation files parsed successfully
- 6 REST deployment option tests passed
- 12 existing deployment utility tests passed


nvuillam:
restDeploy had some bugs years ago that's why i removed it, but you can activate it by default :)

SalikLP:
When you say use it as default, do you mean that we invert the option to useSOAPDeploy, and default to REST or would you rather that people don't have to set anything? Another option is to bake it into the docker image?

nvuillam:
Yes, use REST by default, except if the variable says otherwise, in that case we fallback to SOAP ;)

